PR #2, opened 2019-01-14
initial belga image search

petrjasek:
SDBELGA-9 SDBELGA-7

### DIFF
--- a/client/superdesk.config.js
+++ b/client/superdesk.config.js
@@ -13,6 +13,7 @@ module.exports = function(grunt) {
         ],
         importApps: [
             'superdesk-analytics',
+            'superdesk-planning',
             '../belga',
         ],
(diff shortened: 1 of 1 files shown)


client/superdesk.config.js
```diff
@@ -8,21 +8,15 @@ module.exports = function(grunt) {
...
         apps: [
             'superdesk.analytics',
-            'superdesk-planning'
+            'superdesk-planning',
+            'belga',
         ],
         importApps: [
             'superdesk-analytics',
-            'superdesk-planning'
+            '../belga',
```
ride90: There is a js error:

screenshot 2019-01-14 at 14 51 26


I think it's because of `superdesk-planning` was removed from `importApps`.

PR #3, opened 2019-01-14
[VOCABULARIES] Added vocabularies coming from legacy Belga system

jerome-poisson:
SDBELGA-24

### DIFF
--- a/server/data/vocabularies.json
+++ b/server/data/vocabularies.json
@@ -1,0 +1,1889 @@
+[
+    {
+        "_id": "countries",
+        "schema": {
+            "name": {},
+            "qcode": {}
+        },
+        "type": "manageable",
+        "display_name": "Countries",
+        "unique_field": "qcode",
+        "items": [
+            {
+                "qcode": "AFG",
+                "name": "Afghanistan",
+                "is_active": true
+            },
+            {
+                "qcode": "ALA",
+                "name": "Aland Islands",
+                "is_active": true
+            },
+            {
+                "qcode": "ALB",
+                "name": "Albania",
+                "is_active": true
+            },
+            {
+                "qcode": "DZA",
+                "name": "Algeria",
+                "is_active": true
+            },
+            {
+                "qcode": "ASM",
+                "name": "American Samoa",
+                "is_active": true
+            },
+            {
+                "qcode": "AND",
+                "name": "Andorra",
+                "is_active": true
+            },
+            {
+                "qcode": "AGO",
+                "name": "Angola",
+                "is_active": true
+            },
+            {
+                "qcode": "AIA",
+                "name": "Anguilla",
+                "is_active": true
+            },
+            {
+                "qcode": "ATA",
+                "name": "Antarctica",
+                "is_active": true
+            },
+            {
+                "qcode": "ATG",
+                "name": "Antigua and Barbuda",
+                "is_active": true
+            },
+            {
+                "qcode": "ARG",
+                "name": "Argentina",
+                "is_active": true
+            },
+            {
+                "qcode": "ARM",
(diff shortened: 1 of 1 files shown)


petrjasek:
I think it needs [service](https://github.com/superdesk/superdesk-ntb/blob/master/server/data/vocabularies.json#L87) field in order to be visible in content profile editor

server/data/vocabularies.json
```diff
@@ -0,0 +1,2246 @@
...
+                "is_active": true
+            },
+            {
+                "qcode": "VEN",
+                "name": "Venezuela (Bolivarian Republic)",
+                "is_active": true
+            },
+            {
+                "qcode": "VNM",
+                "name": "Viet Nam",
```
ride90: I think it should be just a Vietnam, not a Viet Nam.

server/data/vocabularies.json
```diff
@@ -0,0 +1,2246 @@
...
+    {
+        "_id": "countries",
+        "schema": {
+            "name": {},
+            "qcode": {}
+        },
+        "type": "manageable",
+        "display_name": "Countries",
+        "unique_field": "qcode",
+        "items": [
```
ride90: @karelpetrak told that it's better to put Belgium on top of the list. 
I'm not sure if it will change an order in the UI :)

server/data/vocabularies.json
```diff
@@ -0,0 +1,2246 @@
...
+            },
+            {
+                "name": "Zip",
+                "qcode": "Zip",
+                "is_active": true
+            }
+        ]
+    },
+    {
+        "_id": "asset_types",
```
ride90: I'm not sure whether we need `asset_types` cv at all.

server/data/vocabularies.json
```diff
@@ -0,0 +1,2246 @@
...
+            },
+            {
+                "qcode": "ZWE",
+                "name": "Zimbabwe",
+                "is_active": true
+            }
+        ]
+    },
+    {
+        "_id": "asset_formats",
```
ride90: I think that we don't need `asset_formats`. 
Please, clarify it with @karelpetrak

server/data/vocabularies.json
jerome-poisson: hum, it's a copy/paste from AP, so it needs to be fixed there too, thx for the catch.

server/data/vocabularies.json
jerome-poisson: checked with @karelpetrak it not needed

server/data/vocabularies.json
jerome-poisson: indeed, checked with @karelpetrak and will remove them, thanks

server/data/vocabularies.json
jerome-poisson: same as above

karelpetrak: Yes, agree with @ride90. The list of asset formats reflects the way the data are stored in Belga's database now. Users will not interact with it in SD. Thus we can leave them out.

server/data/vocabularies.json
```diff
@@ -0,0 +1,2171 @@
...
+            },
+            {
+                "name": "SPN",
+                "qcode": "SPN",
+                "is_active": true
+            }
+        ]
+    },
+    {
+        "_id": "urgency",
```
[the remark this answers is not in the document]
jerome-poisson: heck I thought I did :-/. Ok after checking, I've made a typo in my mapping file.
About urgency, we need it, but that's right that default values are the same, so I can skip this one too.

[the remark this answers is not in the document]
jerome-poisson: @ride90 updated thanks

PR #4, opened 2019-01-18
rename config

tomaskikutis:
REQUIRED FOR: superdesk/superdesk-client-core#2775


tomaskikutis:
You shouldn't have merged it yet since the client PR is still in review.

PR #14, opened 2019-01-28
Add categories and subscriber_types CVs.

ride90:


server/data/vocabularies.json
```diff
@@ -2392,5 +2392,246 @@
...
+                    }
+                ]
+            }
+        ]
+    },
+    {
+        "_id": "categories",
+        "display_name": "Categories",
+        "type": "manageable",
+        "selection_type": "multi selection",
```
petrjasek: imo this might need service field in order to work in content profile editor

PR #21, opened 2019-03-11
SDBELGA-30 Convert text to html for body_html field.

ride90:


server/belga/io/feed_parsers/base_belga_newsml_1_2.py
```diff
@@ -663,3 +663,12 @@ def is_not_formatted(qcode):
...
         return formatted_subjects
+
+    def _plain_to_html(self, text):
+        # handle newline
+        for newline in ('\r\n', '\n', '\r'):
+            text = text.replace(newline, '</p><p>')
+        # remove redundant whitespaces
+        text = ' '.join(text.split())
+
+        return '<p>' + text + '</p>'
```
jerome-poisson: if you convert plain text to html, I guess you need to escape the text, what if a `<` or `>` is present else?
Also it would probably make sense to move this method to some generic place so it can be re-used, for instance in `superdesk/etree.py`

server/belga/io/feed_parsers/base_belga_newsml_1_2.py
```diff
@@ -663,3 +664,13 @@ def is_not_formatted(qcode):
...
 
         return formatted_subjects
+
+    def _plain_to_html(self, text):
+        # handle newline
+        for newline in ('\r\n', '\n', '\r'):
+            text = text.replace(newline, '</p><p>')
+        # remove redundant whitespaces
+        text = ' '.join(text.split())
+        html.escape(text)
```
jerome-poisson: if you escape here, you'll escape the `</p><p>` you've done just above

PR #23, opened 2019-03-14
update publish queue expiry, multimedia validation

karelpetrak:
set publish queue expiry to 10 days
changed fields for media validation


server/settings.py
```diff
@@ -87,23 +87,20 @@
...
         "required": True,
     },
-    "copyrightholder": {
-        "required": False,
-    },
     "byline": {
         "required": False,
     },
+     "copyrightnotice": {
+     "required": False,
```
petrjasek: indentation is wrong here and on previous line too

PR #25, opened 2019-03-18
added schema and editor config for multimedia

karelpetrak:
added schema and editor config for pictures, video, audio
updated config for VALIDATOR_MEDIA_METADATA

### DIFF
--- a/server/settings.py
+++ b/server/settings.py
@@ -89,21 +89,96 @@ ORGANIZATION_NAME_ABBREVIATION = env('ORGANIZATION_NAME_ABBREVIATION', 'Belga')
 
 PUBLISH_QUEUE_EXPIRY_MINUTES = 60 * 24 * 30  # 30d
 
-# media required fields
+# schema for images
+SCHEMA = {
+    'picture': {
+        'keywords': {'required': False},
+        'ednote': {'required': False},
+        'headline': {'required': False},
+        'description_text': {'required': True},
+        'byline': {'required': False},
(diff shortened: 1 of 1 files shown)


server/settings.py
```diff
@@ -89,21 +89,96 @@
...
+        'headline': {'order': 3, 'sdWidth': 'full'},
+        'description_text': {'order': 4, 'sdWidth': 'full'},
+        'byline': {'order': 5, 'sdWidth': 'half'},
+        'copyrightnotice': {'order': 6, 'sdWidth': 'half'},
+        'sign_off': {'order': 7, 'sdWidth': 'half'},
+    },
+}
+
+# schema for video
+SCHEMA = {
```
petrjasek: you should not redefine SCHEMA again, but rather put all schemas together:
```
SCHEMA = {
    'picture': {
        ...
    },
    'video': {
        ...
    },
    'audio': {
        ...
    },
}
```

server/settings.py
```diff
@@ -89,21 +89,96 @@
...
+        'headline': {'required': False},
+        'description_text': {'required': True},
+        'byline': {'required': False},
+        'copyrightnotice': {'required': False},
+        'sign_off': {'required': False},
+    },
+}
+
+# editor for video
+EDITOR = {
```
petrjasek: same with editor

karelpetrak: fixed, pls check

PR #38, opened 2019-04-08
[SDBELGA-87] Import Media contacts from Belga Box to Superdesk

idsvn:
Create a command (contact:import) to import the contact file (json format).
Add unit test.


server/belga/command/contacts_import.py
```diff
@@ -0,0 +1,151 @@
...
+
+class ContactImportCommand(superdesk.Command):
+    """Import contact from belga to Superdesk.
+    This command use for inserting a large number contact from Belga to Superdesk.
+    Only support for format json file.
+    """
+
+    option_list = [
+        superdesk.Option('--file', '-f', dest='contacts_file_path',
+                         default='/home/thanhnguyen/workspace/0_test/contact/contacts.json')
```
petrjasek: I guess there should won't be any default value, or maybe some filename in local folder

server/settings.py
```diff
@@ -24,7 +23,8 @@
     'apps.languages',
     'planning',
     'belga.image',
-    'belga.io'
+    'belga.io',
+    'belga.command'
```
petrjasek: you can add `,` to simplify the diff

server/belga/command/contacts_import.py
idsvn: yes, I forget it. Thanks

PR #42, opened 2019-04-25
set template for planning export as article

petrjasek:
SDBELGA-104


petrjasek:
requires https://github.com/superdesk/superdesk-planning/pull/1185

PR #47, opened 2019-05-28
implement custom field type belga coverage

petrjasek:
SDBELGA-86 SDBELGA-8


client/belga/image/belga-coverage-images.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+        const className = this.props.rendition === 'thumbnail' ?
+            'flex-grid flex-grid--boxed flex-grid--wrap-items flex-grid--small-3' :
+            'flex-grid flex-grid--boxed flex-grid--wrap-items flex-grid--small-1';
+
+        return (
+            <div className={className}>
+                {this.state.images.map((image) => (
+                    <div key={image.imageId} className="flex-grid__item">
+                        <img src={image[this.props.rendition + 'Url']} alt={image.name} />
```
tomaskikutis: Don't use dynamic fields like `image[this.props.rendition + 'Url']`. It's not possible to find usages of field names this way. I prefer writing a helper function. It is longer but more explicit and easier to understand and maintain.
```
function getUrl(rendition: 'preview' | 'thumbnail') {
    if (rendition === 'preview') {
        return image['previewUrl'];
    } else if (rendition === 'thumbnail') {
        return image['thumbnailUrl'];
    } else {
        assertNever(rendition);
    }
}
```

client/belga/image/belga-coverage-editor.tsx
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
+import {gettext} from 'superdesk-core/scripts/core/utils';
+import {DropZone} from 'superdesk-core/scripts/core/ui/components/drop-zone';
+import {IEditorComponentProps} from 'superdesk-core/scripts/apps/fields';
+import BelgaCoverageInfo from './belga-coverage-info';
+import BelgaCoverageImages from './belga-coverage-images';
+
+const ALLOWED_TYPES = [
+    'application/superdesk.item.graphic',
```
tomaskikutis: Which common functionality do we need from Superdesk core? I'd like the client repositories to be as independent as possible and it would be good if we could not depend on things where it can be avoided. Can we use some custom type specific to coverages?

client/belga/image/belga-coverage-info.tsx
```diff
@@ -0,0 +1,73 @@
...
+            return this.renderCoverage(this.state.coverage);
+        }
+
+        return <div>{'There was an error when fetching coverage.'}</div>;
+    }
+
+    renderCoverage(coverage: IBelgaCoverage) {
+        const inEditor = this.props.removeCoverage != null;
+        return (
+            <div className={inEditor ? "sd-media-carousel__content" : ''}>
```
tomaskikutis: I came up with an idea to solve CSS interdependencies. Belga coverages shouldn't depend on CSS used by other features in client-core. How about if we create a React UI component, say Carousel, export it from the extensions API, use it here like

```
<Carousel>
    <CarouselItem caption={'test'}>
        <img />
    </CarouselItem>
    <CarouselItem>
        <img />
    </CarouselItem>
</Carousel>
```

The implementation would still use CSS classes related to media carousel/associations, but we could then hand this over to the UI team to refactor the CSS and move it to UI framework when it's done. What do you think?

client/belga/image/belga-coverage-editor.tsx
```diff
@@ -0,0 +1,47 @@
...
+                    <BelgaCoverageImages
+                        coverageId={this.props.value}
+                        rendition={'thumbnail'}
+                        maxImages={3}
+                    />
+                </div>
+            );
+        }
+
+        return this.renderDropzone();
```
tomaskikutis: Can you inline this too?

client/belga/image/belga-coverage-info.tsx
petrjasek: yep that would be ideal, will add some core components for these

client/belga/image/belga-image-api.ts
```diff
@@ -0,0 +1,36 @@
...
+    previewUrl: string;
+    thumbnailUrl: string;
+}
+
+const API_URL = 'https://api.ssl.belga.be/belgaimage-api/';
+
+const callApi = (endpoint, params) => {
+    const queryString = Object.keys(params).map((key) => `${key}=${params[key]}`).join('&');
+
+    return fetch(API_URL + endpoint + '?' + queryString)
```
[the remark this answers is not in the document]
petrjasek: there is no auth

client/extensions/belgaCoverage/src/belga-coverage-info.tsx
```diff
@@ -0,0 +1,53 @@
...
+        if (this.state.loading) {
+            return null;
+        }
+
+        if (this.state.coverage === null) {
+            return <Alert type="error">{'There was an error when fetching coverage.'}</Alert>;
+        }
+
+        // workaround for this.state.coverage can be null
+        const coverage: IBelgaCoverage = this.state.coverage!;
```
tomaskikutis: [Here's](https://user-images.githubusercontent.com/2076953/59294878-c0cfed00-8c82-11e9-83a3-6c2b448bc899.png)  how to avoid the need for a workaround. TypeScript has a hard time inferring types when using class property initializers, so I'd rather stick with constructors.

client/extensions/belgaCoverage/src/belga-coverage-images.tsx
```diff
@@ -0,0 +1,73 @@
...
+        case 'preview':
+            return image.previewUrl;
+        case 'thumbnail':
+            return image.thumbnailUrl;
+    }
+}
+
+export default class BelgaCoverage extends React.Component<IProps, IState> {
+
+    readonly state = {loading: false, images: [] as IBelgaImage[]};
```
tomaskikutis: Replace class property initializer with a constructor and TypeScript will infer the type from the state signature correctly without having to use `as` syntax which should be avoided when possible.

client/extensions/belgaCoverage/src/belga-coverage-editor.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+    const {gettext} = props.superdesk.localization;
+    const {DropZone} = props.superdesk.components;
+
+    return (
+        <DropZone
+            label={gettext('Drop coverage here')}
+            canDrop={isAllowedType}
+            onDrop={(event) => {
+                const coverage = JSON.parse(getData(event));
```
tomaskikutis: Use a `try-catch` when parsing JSON, it may throw an error.

client/extensions/belgaCoverage/src/belga-coverage-editor.tsx
```diff
@@ -13,37 +13,39 @@ function getData(event: DragEvent) {
...
-                    coverageId={props.value}
-                    rendition={'thumbnail'}
-                    maxImages={3}
-                    superdesk={props.superdesk}
-                />
-            </div>
-        );
-    }
+export function getBelgaCoverageEditor(superdesk: ISuperdesk): React.StatelessComponent<IEditorComponentProps> {
+    return function BelgaCoverageEditor(props: IEditorComponentProps) {
```
tomaskikutis: You can also use `return class A extends React.PureComponent {...}`.

client/extensions/belgaCoverage/src/index.ts
```diff
@@ -0,0 +1,27 @@
+import {ISuperdesk, IExtension, IExtensionActivationResult} from 'superdesk-api';
```
tomaskikutis: If you committed a rename without any changes and added changes in another commit I think git would be able to 
 preserve the diff.

client/belga/image/belga-coverage-editor.tsx
petrjasek: it gets belga images from search provider where this mimetype is set

client/belga/image/belga-coverage-info.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+export default class BelgaCoverageAssocation extends React.Component<IProps, IState> {
+    readonly state = {loading: true, coverage: null};
+
+    componentDidMount() {
+        this.fetchCoverage();
+    }
+
+    componentDidUpdate(prevProps) {
+        if (prevProps.coverageId !== this.props.coverageId) {
```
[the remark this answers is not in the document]
petrjasek: it wouldn't fetch new data from server, so this part is still needed

client/extensions/belgaCoverage/src/belga-coverage-editor.tsx
petrjasek: is there any benefit when doing it like that?

tomaskikutis: That's the same benefits when using functional vs class components. Functional ones can't have refs or lifecycle methods. Pure component automatically implements `shoudComponentUpdate` (implements referential checking of props) and will perform better without much effort.

petrjasek: isn't it more explicit then if it's not using any of those features?

tomaskikutis: You can keep it if it's clearer for you. `React.Component` / `React.PureComponent` are basic building blocks so I don't see them as causing any issues.

petrjasek: yep it's not causing any issues, I was just wondering if there is any typescript/etc benefits I'm not aware of

client/belga/image/belga-coverage-info.tsx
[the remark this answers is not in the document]
tomaskikutis: Did you try it or is this a guess? If you use `props.coverageId` as the key when rendering this component and `coverageId` changes, React will unmount the component and mount it again, which means `componentDidMount` will fire again and correct data would be fetched from the server.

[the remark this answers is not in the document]
petrjasek: I did try and it didn't work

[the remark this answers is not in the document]
tomaskikutis: Hm. You might have placed the key in the wrong place. Check this example https://jsfiddle.net/7yLwfznm/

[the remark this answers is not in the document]
petrjasek: ah stupid me, I did it in the wrong place, works fine, will update

PR #54, opened 2019-07-08
[ SDBELGA-130] Add email attachment for Belga

idsvn:
Register new feeding service Email Belga.
Add function email attachment:
+ get attachment from email.
+ save file to storage and attachments collection.
+ add value for 'ednote' property and save attachment id to ingest item

### DIFF
--- a/server/belga/io/feeding_services/email_belga.py
+++ b/server/belga/io/feeding_services/email_belga.py
@@ -11,6 +11,7 @@ import socket
 import email
 import imaplib
 import io
+import logging
 from flask import current_app as app
 from superdesk.media.media_operations import process_file_from_stream
 from superdesk.io.feeding_services import EmailFeedingService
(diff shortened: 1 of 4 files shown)


server/belga/io/feeding_services/email_belga.py
```diff
@@ -0,0 +1,154 @@
...
+                            file_name, content_type, metadata = res
+                            content.seek(0)
+                            media_id = app.media.put(content,
+                                                     filename=fileName,
+                                                     content_type=content_type,
+                                                     metadata=metadata,
+                                                     resource='attachments')
+                            attachment_service = get_resource_service('attachments')
+                            ids = attachment_service.post([{"media": media_id,
+                                                            "title": 'attachment',
```
petrjasek: maybe it should use the filename here, so you can differentiate in the list

PR #70, opened 2019-10-07
[SDBELGA-172] Tip ingest

idsvn:
- add new feed parser Belga tip format
- add unit test

### DIFF
--- a/server/belga/io/feed_parsers/belga_tip_newsml_1_2.py
+++ b/server/belga/io/feed_parsers/belga_tip_newsml_1_2.py
@@ -1,0 +1,166 @@
+import logging
+from .belga_newsml_1_2 import BelgaNewsMLOneFeedParser
+from superdesk.io.registry import register_feed_parser
+from superdesk.publish.formatters.newsml_g2_formatter import XML_LANG
+from .base_belga_newsml_1_2 import SkipItemException
+logger = logging.getLogger(__name__)
+
+
+class BelgaTipNewsMLOneFeedParser(BelgaNewsMLOneFeedParser):
+    """Feed Parser which can parse specific Belga News ML xml files."""
+
+    NAME = 'belgatipnewsml12'
+
+    label = 'Belga Tip News ML 1.2 Parser'
+
+    SUPPORTED_ASSET_TYPES = ('TIP')
+
+    def parser_newsitem(self, newsitem_el):
+        """
+        Parse Newsitem element.
+
+        Example:
+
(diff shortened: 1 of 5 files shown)


server/belga/io/feed_parsers/belga_tip_newsml_1_2.py
```diff
@@ -0,0 +1,166 @@
...
+
+
+class BelgaTipNewsMLOneFeedParser(BelgaNewsMLOneFeedParser):
+    """Feed Parser which can parse specific Belga News ML xml files."""
+
+    NAME = 'belgatipnewsml12'
+
+    label = 'Belga Tip News ML 1.2 Parser'
+
+    SUPPORTED_ASSET_TYPES = ('TIP')
```
ride90: It must be a `('TIP',)` otherwise this statement: 
```
if not (role_name and role_name.upper() in self.SUPPORTED_ASSET_TYPES)
``` 
is equal to this 
```
if not (role_name and role_name.upper() in ['T', 'I', 'P']
```

server/belga/io/feed_parsers/belga_tip_newsml_1_2.py
```diff
@@ -0,0 +1,166 @@
...
+            </Identification>
+            <NewsManagement>
+                ....
+            </NewsManagement>
+            <NewsComponent>
+                ....
+            </NewsComponent>
+          </NewsItem>
+
+        :param item:
```
ride90: Where it comes from? I think the same is in `BelgaNewsMLOneFeedParser`. A correct docstring only in `BaseBelgaNewsMLOneFeedParser`. It would be good to adjust it.

server/belga/io/feed_parsers/belga_tip_newsml_1_2.py
```diff
@@ -0,0 +1,166 @@
...
+        self.parser_descriptivemetadata(item, descript_el)
+
+        # get 3rd level NewsComponent
+        # body_html, headline, abstract
+        for formalname, item_key in (('Body', 'body_html'), ('Title', 'headline'), ('Lead', 'abstract')):
+            role = newscomponent_el.find('NewsComponent/Role[@FormalName="{}"]'.format(formalname))
+            if role is not None:
+                newscomponent = role.getparent()
+                datacontent = newscomponent.find('ContentItem/DataContent')
+                format = newscomponent.find('ContentItem/Format')
```
ride90: this name is reserved, use `_format` instead

PR #74, opened 2019-10-11
Use sign off as avatar

tomaskikutis:
SDBELGA-149


client/belga/index.tsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import angular from 'angular';
+import {IUser} from 'superdesk-api'
+import {startApp} from 'superdesk-core/scripts/index';
+import belgaImage from './image';
+
+class UserAvatar extends React.Component<{user: IUser}> {
```
petrjasek: what about using PureComponent?

PR #76, opened 2019-10-17
[SDBELGA-187] Ingest from Belga Remote

idsvn:
Add type `ORIGINAL` for SUPPORTED_ASSET_TYPES.
Unit test for belga_remote.

### DIFF
--- a/server/tests/io/feed_parsers/belga_newsml_1_2_test.py
+++ b/server/tests/io/feed_parsers/belga_newsml_1_2_test.py
@@ -19,15 +19,21 @@ from belga.io.feed_parsers.belga_newsml_1_2 import BelgaNewsMLOneFeedParser
 
 class BelgaNewsMLOneTestCase(TestCase):
     filename = 'belga_newsml_1_2.xml'
+    filename_belga_remote = 'belga_remote_newsml_1_2.xml'
 
     def setUp(self):
         dirname = os.path.dirname(os.path.realpath(__file__))
         fixture = os.path.normpath(os.path.join(dirname, '../fixtures', self.filename))
         provider = {'name': 'test'}
+        parser = BelgaNewsMLOneFeedParser()
         with open(fixture, 'rb') as f:
-            parser = BelgaNewsMLOneFeedParser()
             self.xml_root = etree.parse(f).getroot()
             self.item = parser.parse(self.xml_root, provider)
+        # belga remote
+        fixture = os.path.normpath(os.path.join(dirname, '../fixtures', self.filename_belga_remote))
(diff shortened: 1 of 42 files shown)


server/tests/io/feed_parsers/belga_newsml_1_2_test.py
```diff
@@ -19,15 +19,21 @@
 
 class BelgaNewsMLOneTestCase(TestCase):
     filename = 'belga_newsml_1_2.xml'
+    filename_belga_remote = 'belga_remote_newsml_1_2.xml'
```
ride90: Just a small note: I think it's better to create a new 
```
BelgaRemoteNewsMLOneTestCase(BelgaNewsMLOneTestCase)
``` 
class and inherit it from `BelgaNewsMLOneTestCase`, so you can simply overwrite `filename` with the file you want to use. For now, it's fine, but if we need to add more tests in the future, the current approach will slow down things. For example, redundant file reading, parsing and etc is happening when `test_can_parse` and `test_content` run. We don't really need to read and parse `belga_remote_newsml_1_2.xml` if we test only parsed data from `belga_newsml_1_2.xml`.
It's 5 mins fix, but IMO is better to make it.

server/tests/io/fixtures/belga_remote_newsml_1_2.xml
```diff
@@ -0,0 +1,133 @@
...
+        </NewsManagement>
+        <NewsComponent Duid="0" xml:lang="nl">
+            <NewsLines>
+                <HeadLine>Knooppunt Schijnpoort in Antwerpen hele zomervakantie afgesloten</HeadLine>
+            </NewsLines>
+            <AdministrativeMetadata/>
+            <DescriptiveMetadata>
+                <Genre FormalName=""/>
+            </DescriptiveMetadata>
+            <NewsComponent Duid="0" xml:lang="nl">
```
ride90: I wonder if all `belga remote` files have `Duid="0"`, if so, we must adjust a logic. Please, clarify this moment.

idsvn: Hi, @ride90 
IMO, In `Belga remote` we should use `dateid`(instead of `Duid`) for `guid`. and I will split the `Belga remote` to another feed parser. Is is ok?

ride90: Do you think `dateid` is unique? For me, it looks like it's not unique. Can you clarify why `Duid` is 0 with Stefaan?

P.S. If custom logic is needed let's move Belga Remote to a standalone parser.

PR #81, opened 2019-11-01
SDBELGA-202

idsvn:
Update Ingest mapping from AFP,  ANP, TASS, EFE, ATS, KYODO,  DPA:
- Service is always equal NEWS
- Product have only one value (POLITICS, ECONOMY, SPORTS, GENERAL ...)

### DIFF
--- a/server/belga/io/feed_parsers/base_belga_newsml_1_2.py
+++ b/server/belga/io/feed_parsers/base_belga_newsml_1_2.py
@@ -70,6 +70,13 @@ class BaseBelgaNewsMLOneFeedParser(NewsMLOneFeedParser):
                 try:
                     item = item_envelop.copy()
                     self.parser_newsitem(item, newsitem_el)
+                    # add product is GENERAL, if product is empty
+                    if not [it for it in item.get('subject', []) if it.get('scheme') == 'news_products']:
+                        product = {"name": 'GENERAL', "qcode": 'GENERAL', "scheme": "news_products"}
+                        item.setdefault('subject', []).append(product)
+                    # add service is NEW
+                    service = {"name": 'NEWS', "qcode": 'NEWS', "scheme": "news_services"}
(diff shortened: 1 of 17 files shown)


server/belga/io/feed_parsers/base_belga_newsml_1_2.py
```diff
@@ -104,23 +104,6 @@ def parser_newsenvelop(self, envelop_el):
...
-
-        for element in envelop_el.findall('NewsProduct'):
-            if element is not None and element.get('FormalName'):
-                item.setdefault('subject', []).append({
-                    "name": element.get('FormalName'),
-                    "qcode": element.get('FormalName'),
-                    "scheme": "news_products"
-                })
-
         element = envelop_el.find('Priority')
```
idsvn: Belga don't use product and service that get from external source. Instead of this, they will auto mapping the product and service follow key-works, catalog, ....
So we remove these code.

server/belga/io/feed_parsers/belga_anp_newsml_1_2.py
```diff
@@ -32,22 +32,21 @@ def parser_newsmanagement(self, item, manage_el):
...
+        product = {}
+        for subject in item.get('subject', []):
+            if subject.get('scheme', '') == 'genre':
+                qcode = self.MAPPING_PRODUCTS.get(subject.get('name'))
+                product = {'name': qcode, 'qcode': qcode, 'scheme': 'news_products'} if qcode else None
+                break
+        if not product:
+            product = {'name': 'GENERAL', 'qcode': 'GENERAL', 'scheme': 'news_products'}
+        item.setdefault('subject', []).append(product)
+        # service is always equal NEWS
```
ride90: Does it make sense to put it into `BaseBelgaNewsMLOneFeedParser`? 
I see that we perform this action in afp, anp, anpa, efe, iptc7901, tass and so on..

PR #82, opened 2019-11-05
change extensions setup

tomaskikutis:
DEPENDS ON: https://github.com/superdesk/superdesk-client-core/pull/3188
DEPENDS ON: https://github.com/superdesk/superdesk-planning/pull/1383


client/belga/index.tsx
```diff
@@ -4,6 +4,9 @@ import {IUser} from 'superdesk-api'
 import {startApp} from 'superdesk-core/scripts/index';
 import belgaImage from './image';
 import belga360Archive from './360archive';
+import planningExtension from '../node_modules/superdesk-planning/client/planning-extension/dist/src/extension';
+import markForUserExtension from '../node_modules/superdesk-core/scripts/extensions/markForUser/dist/src/extension';
```
petrjasek: just wondering - wouldn't import from `'superdesk-core/scripts/extensions/markForUser'` work?

tomaskikutis: yep, it does work, I'll use that :)

client/package.json
```diff
@@ -1,9 +1,12 @@
 {
     "name": "superdesk",
     "license": "GPL-3.0",
+    "scripts": {
+        "postinstall": "cd extensions/belgaCoverage && npm install && npm run compile"
```
petrjasek: I think it would be better to have this in belgaCoverage package.json file, here add belgaCoverage to dependencies:
```js
"belga-coverage": "file:./extensions/belgaCoverage",
```
and make the import in index like `import belgaCoverageExtension from 'belga-coverage'`

PR #87, opened 2019-11-13
Create macro to add prefix BELG to headline. [SDBELGA-196]

GyanP:
belg-headline

### DIFF
--- a/server/belga/macros/__init__.py
+++ b/server/belga/macros/__init__.py
@@ -5,32 +5,4 @@ This module will get reloaded per request to be up to date.
 Use `superdesk.macro_register.macros.register` for registration.
 """
 
-import os
-import sys
-import imp
-import importlib
-
-
-def load_macros(path):
-    """Load macros from given path
-
-    :param str path:
-    """
-    macros = [f[:-3] for f in os.listdir(path)
-              if f.endswith('.py') and not f.endswith('_test.py') and not f.startswith('__')]
-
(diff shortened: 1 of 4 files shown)


server/belga/macros/__init__.py
```diff
@@ -0,0 +1,36 @@
...
+Use `superdesk.macro_register.macros.register` for registration.
+"""
+
+import os
+import sys
+import imp
+import importlib
+
+
+def load_macros(path):
```
petrjasek: I think you can just import it:
```
import belga.macros.add_prefix_to_headline
```

GyanP: I thought let's keep it consistent with other repo.

server/belga/macros/__init__.py
```diff
@@ -0,0 +1,8 @@
+"""Here you can import/implement macros.
+
+This module will get reloaded per request to be up to date.
+
+Use `superdesk.macro_register.macros.register` for registration.
+"""
```
petrjasek: I think this comment is off now

GyanP:
@petrjasek all other macros from the core like `POPULATE ABSTRACT`, `VALIDATE FOR PUBLISH` etc. are not available anymore. Is that ok?

petrjasek:
@karelpetrak do they need any core macros?

GyanP:
All of them will be available now, as Jerome made some changes.

PR #92, opened 2019-11-18
SDBELGA-152 Implement OpenID Connect authentication

idsvn:
Belga uses KeyCloak server for Single Sign On.


server/requirements.txt
```diff
@@ -7,3 +7,5 @@ oauth2client==4.1.3
 -e git+git://github.com/superdesk/superdesk-core.git@develop#egg=Superdesk-Core
 -e git+git://github.com/superdesk/superdesk-analytics.git@master#egg=superdesk-analytics
 -e git+git://github.com/superdesk/superdesk-planning.git@master#egg=superdesk-planning
+
+flask-oidc==1.4.0
```
idsvn: When we put `flask-oidc` at the begin of file, I conflict with some library in Superdesk-Core and cannot install, So we add it in the end of file.

server/belga/auth/__init__.py
```diff
@@ -0,0 +1,79 @@
...
+class AuthResource(Resource):
+    schema = {
+        'token': {
+            'type': 'string'
+        },
+        'user': Resource.rel('users', True)
+    }
+    resource_methods = ['POST']
+    item_methods = ['GET', 'DELETE']
+    public_methods = ['POST', 'DELETE']
```
petrjasek: not sure if there should be public delete

server/settings.py
```diff
@@ -221,3 +223,11 @@
 KEYWORDS_ADD_MISSING_ON_PUBLISH = True
 MACROS_MODULE = env('MACROS_MODULE', 'belga.macros')
 
+SECRET_KEY = 'SomethingNotEntirelySecret'
+OIDC_CLIENT_SECRETS = env('OIDC_CLIENT_SECRETS', 'client_secrets.json')
```
petrjasek: would be good to use env variables for all secrets, also for client secret parts:
```py
SECRET_KEY = env('SECRET_KEY')
OIDC_CLIENT_SECRETS = {
    'web': {
        'client_id: env('OIDC_CLIENT_ID'),
        'client_secret': env('OIDC_CLIENT_SECRET'),
    },
}
```
or maybe just define those `OIDC_CLIENT_SECRET` etc and construct the config using these in init_app:
```py
def init_app(app):
    app.config.setdefault('OIDC_CLIENT_SECRETS', {
        'web': {
             'client_secret: app.config.get('OIDC_CLIENT_SECRET'),
             ...,
        },
    })
```

server/client_secrets.json
```diff
@@ -0,0 +1,11 @@
+{
```
petrjasek: this file shouldn't be in git, you can add it to `.gitignore`

server/settings.py
idsvn: Unfortunately, there is no release from commit https://github.com/puiterwijk/flask-oidc/commit/de9584677fbed8f8688eab60145744cb353915ea which support loading `CLIENT_SECRET` from dict, we are considering [Flask-OIDC-ex](https://github.com/larsw/flask-oidc-ex), if it's working fine, we will update accordingly.

idsvn:
Keycloak Auth will be implemented in `superdesk-core` so close this one.

PR #94, opened 2019-11-20
Sync vocabularies

ride90:
- multilingual countries CV (single selection) for 'Country' keyword
- explicitly set `selection_type` for some items


karelpetrak:
Hi @ride90 
looks good to me, 
I think we should get rid of **NTB Event format**
https://github.com/superdesk/superdesk-belga/blob/bfb5fa2ff58647db6138056c7caa2327c80fe481/server/data/vocabularies.json#L7019
https://github.com/superdesk/superdesk-belga/blob/bfb5fa2ff58647db6138056c7caa2327c80fe481/server/data/vocabularies.json#L7062
https://github.com/superdesk/superdesk-belga/blob/bfb5fa2ff58647db6138056c7caa2327c80fe481/server/data/vocabularies.json#L7101

Remove **Categories** which are not used either
https://github.com/superdesk/superdesk-belga/blob/bfb5fa2ff58647db6138056c7caa2327c80fe481/server/data/vocabularies.json#L7121

and should also delete this CV: **Of interest to**
https://github.com/superdesk/superdesk-belga/blob/bfb5fa2ff58647db6138056c7caa2327c80fe481/server/data/vocabularies.json#L7652

PR #100, opened 2019-11-28
[MACROS][SET_DEFAULT_METADATA] copy "services-products" +  added test

jerome-poisson:
SDBELGA-241
SDBELGA-142

### DIFF
--- a/server/belga/macros/set_default_metadata.py
+++ b/server/belga/macros/set_default_metadata.py
@@ -12,7 +12,7 @@ from superdesk import get_resource_service
 
 
 logger = logging.getLogger(__name__)
-SUBJECT_SCHEMES = ('news_services', 'news_products', 'distribution')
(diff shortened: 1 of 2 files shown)


server/belga/macros/set_default_metadata.py
```diff
@@ -12,7 +12,7 @@
 
 
 logger = logging.getLogger(__name__)
-SUBJECT_SCHEMES = ('news_services', 'news_products', 'distribution')
+SUBJECT_SCHEMES = ('services-products', 'news_services', 'news_products', 'distribution')
```
ride90: It's fine for now to keep `'news_services', 'news_products'`, but after ingests/output will be migrated to use only `'services-products'` we should delete `'news_services', 'news_products'`.

jerome-poisson: Yes I would have deleted them, but the ticket description asked to keep them.

PR #102, opened 2019-12-02
Created common macro that run two macros as incoming rule for a desk stage. [SDBELGA-234]

GyanP:


petrjasek:
would rather call it *desk incoming macro* or something so we don't have to rename it when adding some more calls or change something

PR #103, opened 2019-12-02
Added default source for manually created items

karelpetrak:


server/settings.py
```diff
@@ -84,6 +84,9 @@
     'wst': 'America/Indiana/Indianapolis'
 }
 
+#: Default value for Source to be set for manually created items
+DEFAULT_SOURCE_VALUE_FOR_MANUAL_ARTICLES = Belga
```
petrjasek: it should be `= "Belga"`

PR #105, opened 2019-12-05
SDBELGA-239 Update mapping of services and products for belga newsml

ride90:


server/data/vocabularies.json
```diff
@@ -5,12 +5,11 @@
         "type": "manageable",
         "selection_type": "multi selection",
         "init_version": 2,
```
petrjasek: you can bump the init version to force reload

ride90: I did not change this CV values, just formatting.

petrjasek: sorry, my bad

PR #108, opened 2019-12-13
SDBELGA-238 Update mapping of services and products in ingest

ride90:


server/belga/io/feed_parsers/base_belga_newsml_1_2.py
```diff
@@ -70,13 +70,17 @@ def parse(self, xml, provider=None):
...
-                        product = {"name": 'GENERAL', "qcode": 'GENERAL', "scheme": "news_products"}
-                        item.setdefault('subject', []).append(product)
+                    # add product is NEWS/GENERAL, if product is empty
+                    if not [it for it in item.get('subject', []) if it.get('scheme') == 'services-products']:
+                        item.setdefault('subject', []).append({
+                            'name': 'NEWS/GENERAL',
+                            'qcode': 'NEWS/GENERAL',
+                            'parent': 'NEWS',
+                            'scheme': 'services-products'
+                        })
```
idsvn: Hi, we see you have a check NEWS/GENERAL in all ingests. Should we need to check it again in here?
It's right. I only see it has little redundant.

ride90: Some ingests don't overload this method, so that's why it's here.

PR #110, opened 2019-12-18
SDBELGA-251update all slugline and keywords is null expect ingests from Belga

idsvn:
Belga want all storytags (it map with keywords in vocabularies) and slugline in all ingest is null expect ingests from Belga.


ride90:
Hi @idsvn , according to the linked ticket:

> 'Slugline' and 'Storytags' to be empty for all items ingested from ANP

current set empty slugline and keywords not only for ANP. Is it correct?

idsvn:
> current set empty slugline and keywords not only for ANP. Is it correct?

Hi @ride90,
We think the ticket needs to be updated, according to the info we got from Knapen Seppe:
> To be clear: storytags and slugline should be empty for AFP, DPA, ANP, ATS, EFE, Itartass and KYODO

PR #111, opened 2019-12-18
Add extension to change urgency and distribution when changing the content profile to or from ALERT. [SDBELGA-230]

GyanP:


client/package.json
```diff
@@ -2,7 +2,7 @@
     "name": "superdesk",
     "license": "GPL-3.0",
     "scripts": {
-        "postinstall": "cd extensions/belgaCoverage && npm install && npm run compile"
+        "postinstall": "node ./tasks/install-extensions.js"
```
tomaskikutis: You shouldn't need to change the setup neither add any additional scripts. Is something not working as is?

client/belga/index.tsx
```diff
@@ -8,6 +8,7 @@ import planningExtension from 'superdesk-planning-extension';
 import markForUserExtension from 'superdesk-core/scripts/extensions/markForUser/dist/src/extension';
 import datetimeFieldExtension from 'superdesk-core/scripts/extensions/datetimeField/dist/src/extension';
 import belgaCoverageExtension from '../extensions/belgaCoverage/dist/index';
+import updateArticleExtension from '../extensions/updateArticle/dist/src/extension';
```
tomaskikutis: Try naming extensions according to features they implement instead of which extension points they use. This one should be probably called `updateArticleOnContentProfileChange` or similar.

client/package.json
GyanP: The `postinstall` script install and compile only the `belgacoverage extension`. Shouldn't it be modified to install and compile all the extensions available inside the extension folder?

tomaskikutis: Sorry, my bad, you do need to compile the new extension. I would have simply added another `cd /extension-path && npm run compile` command to `postinstall`, but a script also works.

ride90:
@GyanP what's the status of this PR?

GyanP:
Hi @ride90, it has a dependency on the client-core PR, so, the build of this PR will pass only after that PR gets merged. And that PR is waiting for review from Tomas.

PR #114, opened 2019-12-30
[SDBELGA-242] Add slugline for belga 360 search item

idsvn:
Changes:
- Map field `topic` from belga 360 to `slugline` in superdesk

### DIFF
--- a/server/belga/search_providers.py
+++ b/server/belga/search_providers.py
@@ -281,6 +281,7 @@ class Belga360ArchiveSearchProvider(superdesk.SearchProvider):
             '_id': guid,
             'guid': guid,
             'headline': get_text(data['headLine']),
+            'slugline': get_text(data['topic']),
             'name': get_text(data['name']),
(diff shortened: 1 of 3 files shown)


server/belga/search_providers.py
```diff
@@ -281,6 +281,7 @@ def format_list_item(self, data):
             '_id': guid,
             'guid': guid,
             'headline': get_text(data['headLine']),
+            'slugline': get_text(data['topic']),
```
ride90: I hope `topic` (and `headLine`) always exists, otherwise we should use `.get`

idsvn: We saw that response from Belga 360 is really consistency, if these `topic`, `headLine` does not exist, it will [return null or empty](https://user-images.githubusercontent.com/45254285/71703736-d405fe00-2e08-11ea-9dd6-68c0135d76f0.png) instead.

PR #121, opened 2020-01-09
brief internal destination macro

petrjasek:
- [x] schedule publishing +30m
- [x] update subject
- [x] update header
- [x] update body
- [x] tests on travis


petrjasek:
uses https://github.com/superdesk/superdesk-core/pull/1770

server/.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
ride90: is it ok to commit this?

petrjasek: no

PR #122, opened 2020-01-10
SDBELGA-258 Support for a standalone video item in BelgaNewsML12Formatter.

ride90:


server/tests/publish/belga_newsml_1_2_tests.py
```diff
@@ -2414,3 +2414,387 @@ def test_related_article_belga_archive_text(self):
...
+            {
+                '_id': 'pdf_1',
+                'content': BytesIO(b'it is a pdf'),
+                'content_type': 'application/pdf',
+                'metadata': {
+                    'length': 40248
+                }
+            },
+        )
+        for media_item in media_items:
```
idsvn: We only need to put video here, don't we?

small note: There is extra 1.2 https://github.com/superdesk/superdesk-belga/blob/0b6f39ea9b033e193d4b96d350b9230a26e0caa1/server/belga/publish/belga_newsml_1_2.py#L37

ride90: Yes, it's copy-paste issue :) Only video.

PR #124, opened 2020-01-17
[SDBELGA-266] Add multilingual country for belga ingest item

idsvn:
Add multilingual country support for following NewsML format:
- AFP
- DPA (xml)
- EFE
- Kyoudo

Following format uses two character country code so it will always empty:
- [ANP](https://github.com/idsvn/superdesk-belga/blob/SDBELGA-266/server/tests/io/fixtures/anp_belga.xml#L58)
- [ATS](https://github.com/idsvn/superdesk-belga/blob/SDBELGA-266/server/tests/io/fixtures/ats_newsml_1_2_belga.xml#L64)

For ATS, DPA text, from what we saw, we currently don't parse country
For TASS, if we use [NewsComponent/NewsComponent](https://github.com/idsvn/superdesk-belga/blob/SDBELGA-266/server/belga/io/feed_parsers/belga_tass_newsml_1_2.py#L76), `vocabularies` service for `country` will return None, @ride90 Could you give us your opinion on this?
SDBELGA-266

### DIFF
--- a/server/belga/io/feed_parsers/base_belga_newsml_1_2.py
+++ b/server/belga/io/feed_parsers/base_belga_newsml_1_2.py
@@ -85,8 +85,6 @@ class BaseBelgaNewsMLOneFeedParser(NewsMLOneFeedParser):
                     # Slugline and keywords is epmty
                     item['slugline'] = None
                     item['keywords'] = []
-                    # delete subject is duplicated
-                    item['subject'] = [dict(t) for t in {tuple(d.items()) for d in item['subject']}]
                     item = self.populate_fields(item)
                 except SkipItemException:
                     continue
@@ -557,7 +555,10 @@ class BaseBelgaNewsMLOneFeedParser(NewsMLOneFeedParser):
             elements = location_el.findall('Property')
             for element in elements:
                 if element.attrib.get('FormalName', '') == 'Country':
-                    item['extra']['country'] = element.attrib.get('Value')
+                    country = element.attrib.get('Value')
+                    item['extra']['country'] = country
+                    # country keywords is CV
+                    item.setdefault('subject', []).extend(self._get_country(country))
                 if element.attrib.get('FormalName', '') == 'City':
                     item['extra']['city'] = element.attrib.get('Value')
                 if element.attrib.get('FormalName', '') == 'CountryArea':
@@ -718,3 +719,15 @@ class BaseBelgaNewsMLOneFeedParser(NewsMLOneFeedParser):
 
     def _get_cv(self, _id):
         return superdesk.get_resource_service('vocabularies').find_one(req=None, _id=_id)
+
+    def _get_country(self, country):
+        country_keyword = [
+            c for c in self._get_cv('country').get('items', [])
+            if c.get('qcode') == 'country_' + country.lower() and c.get('is_active')
+        ]
+        # add scheme and remove is_active
+        for c in country_keyword:
+            c['scheme'] = 'country'
+            if 'is_active' in c:
+                c.pop('is_active')
+        return country_keyword
--- a/server/belga/io/feed_parsers/base_belga_newsml_1_2.py
+++ b/server/belga/io/feed_parsers/base_belga_newsml_1_2.py
@@ -558,6 +558,8 @@ class BaseBelgaNewsMLOneFeedParser(NewsMLOneFeedParser):
                 item['extra']['how_present'] = how_present_el
 
             elements = location_el.findall('Property')
+            # avoid repeatedly calling db while finding country
+            self.country = self._get_cv('country').get('items', [])
             for element in elements:
                 if element.attrib.get('FormalName', '') == 'Country':
                     country = element.attrib.get('Value')
@@ -725,10 +727,10 @@ class BaseBelgaNewsMLOneFeedParser(NewsMLOneFeedParser):
     def _get_cv(self, _id):
         return superdesk.get_resource_service('vocabularies').find_one(req=None, _id=_id)
 
-    def _get_country(self, country):
+    def _get_country(self, country_code):
         country_keyword = [
-            c for c in self._get_cv('country').get('items', [])
-            if c.get('qcode') == 'country_' + country.lower() and c.get('is_active')
+            c for c in self.country
+            if c.get('qcode') == 'country_' + country_code.lower() and c.get('is_active')
         ]
         # add scheme and remove is_active
         for c in country_keyword:
--- a/server/belga/io/feed_parsers/base_belga_newsml_1_2.py
+++ b/server/belga/io/feed_parsers/base_belga_newsml_1_2.py
@@ -17,6 +17,8 @@ from superdesk.etree import etree
 from superdesk.io.feed_parsers.newsml_1_2 import NewsMLOneFeedParser
 from superdesk.io.iptc import subject_codes
 
+from .belga_newsml_mixin import BelgaNewsMLMixin
+
 
 class SkipItemException(Exception):
     """Raised when we current item must be skipped"""
@@ -24,12 +26,8 @@ class SkipItemException(Exception):
     pass
 
 
-class BaseBelgaNewsMLOneFeedParser(NewsMLOneFeedParser):
+class BaseBelgaNewsMLOneFeedParser(BelgaNewsMLMixin, NewsMLOneFeedParser):
     """Base Feed Parser for NewsML format, specific AFP, ANP, .. Belga xml."""
-
-    def __init__(self):
-        super().__init__()
-        self._countries = []
 
     def parse(self, xml, provider=None):
         """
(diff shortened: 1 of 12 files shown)


ride90:
@idsvn flake8 issues

server/belga/io/feed_parsers/base_belga_newsml_1_2.py
```diff
@@ -85,8 +85,6 @@ def parse(self, xml, provider=None):
                     # Slugline and keywords is epmty
                     item['slugline'] = None
                     item['keywords'] = []
-                    # delete subject is duplicated
-                    item['subject'] = [dict(t) for t in {tuple(d.items()) for d in item['subject']}]
```
ride90: why it's deleted?

server/belga/io/feed_parsers/base_belga_newsml_1_2.py
```diff
@@ -557,7 +555,10 @@ def parser_descriptivemetadata(self, item, descript_el):
             elements = location_el.findall('Property')
             for element in elements:
                 if element.attrib.get('FormalName', '') == 'Country':
-                    item['extra']['country'] = element.attrib.get('Value')
+                    country = element.attrib.get('Value')
+                    item['extra']['country'] = country
+                    # country keywords is CV
+                    item.setdefault('subject', []).extend(self._get_country(country))
```
ride90: IMO it's better to load all country keywords once using `def _get_country_keywords` and save them into `self`.
It's not critical, but it's better to avoid hitting DB in the loop.

server/belga/io/feed_parsers/belga_dpa_newsml_2_0.py
```diff
@@ -169,6 +170,18 @@ def parse_content_subject(self, tree, item):
                 name_elt = subject_elt.find(self.qname('name'))
                 if name_elt is not None and not item['extra'].get('country'):
                     item.setdefault('extra', {})['country'] = name_elt.text
+                # looking for ISO 3166-1 alpha-3 code
+                for i in subject_elt.findall(self.qname('sameAs')):
+                    code = i.find(self.qname('name')).text
+                    if len(code) == 3:
+                        country_keyword = self._get_country_keywords(code)
```
ride90: same here

server/tests/io/feed_parsers/belga_afp_newsml_1_2_test.py
```diff
@@ -41,6 +41,7 @@ def test_content(self):
             {'name': 'trials', 'qcode': '02008000', 'scheme': 'iptc_subject_codes'},
             {'name': 'AMN-TFG-1=AMW', 'qcode': 'AMN-TFG-1=AMW', 'scheme': 'of_interest_to'},
             {'name': 'EUA-TFG-1=EUA', 'qcode': 'EUA-TFG-1=EUA', 'scheme': 'of_interest_to'},
+            {'name': 'EUA-TFG-1=EUA', 'qcode': 'EUA-TFG-1=EUA', 'scheme': 'of_interest_to'},
```
ride90: I think it's a duplication of a dict above. Why do we need it?

ride90:
@idsvn 
> For TASS, if we use NewsComponent/NewsComponent, vocabularies service for country will return None, @ride90 Could you give us your opinion on this?

I'm not sure if I understand the issue. Can you provide me a TASS files you've used for testing?

server/belga/io/feed_parsers/base_belga_newsml_1_2.py
idsvn: We use set here so it will not work for nested dict such as `country`
e.g `{'name': 'France', 'qcode': 'country_fra', 'scheme': 'country', 'translations': {'name': {'nl': 'Frankrijk', 'fr': 'France'}}}`
but it seems [itertools.groupby](https://docs.python.org/3/library/itertools.html#itertools.groupby) do the work so we have updated it

idsvn:
> @idsvn
> I'm not sure if I understand the issue. Can you provide me a TASS files you've used for testing?

We updated test case for TASS file, can you run it and check what wrong with `vocabularies` service? Thanks

server/tests/io/feed_parsers/belga_afp_newsml_1_2_test.py
idsvn: because we delete code to remove duplicated subject items, fixed in https://github.com/superdesk/superdesk-belga/pull/124/commits/9f9405f10a2c5438ad75e470c4999996f3dea071

server/belga/io/feed_parsers/base_belga_newsml_1_2.py
```diff
@@ -555,9 +558,14 @@ def parser_descriptivemetadata(self, item, descript_el):
                 item['extra']['how_present'] = how_present_el
 
             elements = location_el.findall('Property')
+            # avoid repeatedly calling db while finding country
+            self.country = self._get_cv('country').get('items', [])
```
ride90: `self.countries`

ride90: I think this is not the correct place to preload countries. Theoretically, if `location_el is None` the `self.country` will not be filled, but if you call `_get_country` later it will raise an error because self.country is not even defined.
I would preload countries in the constructor (__init__) or even with lazy load like that:

```
def __init__(self):
    super().__init__()
    self._countries = []

 ....

def _get_country(self, country_code):

    if not self._countries:
        self._countries = self._get_cv('country').get('items', [])

    country_keyword = [
        c for c in self._countries
        if c.get('qcode') == 'country_' + country_code.lower() and c.get('is_active')
    ]
    # add scheme and remove is_active
    for c in country_keyword:
        c['scheme'] = 'country'
        if 'is_active' in c:
            c.pop('is_active')
    return country_keyword
```
⚠️ this snipped is just from my head, it's not tested for sure.

ride90:
@idsvn , I've tested `BelgaTASSNewsMLOneFeedParser` and location is loaded as `'extra': {'city': 'MOSCOW', 'country': 'RUS', 'how_present': 'Origin'},` and as subject:
```
'subject': [{'name': 'NEWS/ECONOMY',
               'parent': 'NEWS',
               'qcode': 'NEWS/ECONOMY',
               'scheme': 'services-products'},
              {'name': 'News', 'qcode': 'News', 'scheme': 'news_item_types'},
              {'name': 'Russian Federation',
               'qcode': 'country_rus',
               'scheme': 'country',
               'translations': {'name': {'fr': 'Russie', 'nl': 'Rusland'}}},
              {'name': 'TASS', 'qcode': 'TASS', 'scheme': 'credits'},
              {'name': 'default', 'qcode': 'default', 'scheme': 'distribution'},
              {'name': 'no', 'qcode': 'no', 'scheme': 'essential'},
              {'name': 'no', 'qcode': 'no', 'scheme': 'equivalents_list'},
              {'name': 'normal', 'qcode': 'normal', 'scheme': 'link_type'}],
``` 
which is ok, isn't it?

server/belga/io/feed_parsers/base_belga_newsml_1_2.py
idsvn: Good point, make sense to do lazy for countries

idsvn:
> which is ok, isn't it?

Yes, that is expected result, but when we run it with `pytest`, it keeps returning `None` for **countries vocabularies**

ride90:
> Yes, that is expected result, but when we run it with pytest, it keeps returning None for countries vocabularies

The reason why it fails is that vocabularies are not preloaded. 
How to fix:
Inherit `BelgaTestCase`:
```
class BelgaTASSNewsMLOneTestCase(BelgaTestCase):
```
and call parent's setup before a main one:
```
    def setUp(self):
        super().setUp()
        dirname = os.path.dirname(os.path.realpath(__file__))
        fixture = os.path.normpath(os.path.join(dirname, '../fixtures', self.filename))
        provider = {'name': 'test'}
        with open(fixture, 'rb') as f:
            parser = BelgaTASSNewsMLOneFeedParser()
            self.xml_root = etree.parse(f).getroot()
            self.item = parser.parse(self.xml_root, provider)
```

idsvn:
> The reason why it fails is that vocabularies are not preloaded.

How could we miss that! Thank you for pointing out, we will fix it on next commit

server/belga/io/feed_parsers/base_belga_newsml_1_2.py
```diff
@@ -718,3 +728,18 @@ def _plain_to_html(self, text):
 
     def _get_cv(self, _id):
         return superdesk.get_resource_service('vocabularies').find_one(req=None, _id=_id)
+
+    def _get_country(self, country_code):
```
petrjasek: I think it would be good to create mixin for this so it's not duplicated

server/belga/io/feed_parsers/base_belga_newsml_1_2.py
```diff
@@ -718,3 +728,18 @@ def _plain_to_html(self, text):
...
+
+        country_keyword = [
+            c for c in self._countries
+            if c.get('qcode') == 'country_' + country_code.lower() and c.get('is_active')
+        ]
+        # add scheme and remove is_active
+        for c in country_keyword:
+            c['scheme'] = 'country'
+            if 'is_active' in c:
+                c.pop('is_active')
```
petrjasek: I think here it modifies the item in `self._countries` so next time `is_active` won't be there
for the previous filter `and c.get('is_active')`. would be better to return a copy to avoid it:
```python
return [{'name': c['name'], 'qcode': c['qcode'], 'scheme': 'country'} for c in country_keyword]
```

PR #125, opened 2020-01-21
Create macro to add 'TIP' to headline and change urgency. [SDBELGA-273]

GyanP:


server/belga/macros/change_headline_and_urgency.py
```diff
@@ -0,0 +1,18 @@
+"""This macro adds a prefix string 'TIP' to headline and sets urgency to 5 of the item"""
+
+
+def change_headline_and_urgency(item, **kwargs):
+    if 'headline' in item:
+        item['headline'] = 'TIP ' + item.get('headline', '')
+
+    if 'urgency' in item:
```
petrjasek: why this condition?

GyanP: Actually, I wasn't sure if we should add the urgency even if the item doesn't have it already.

PR #129, opened 2020-02-03
[SDBELGA-263] Fix ingest error on importing AFP without headline

idsvn:
This issue is related to https://github.com/superdesk/superdesk-belga/pull/65 (SDBELGA-181), where we will use first line of body as its headline.
If we ingest AFP file with the following body, it will fail
```xml
<body.content><p>example</p><p>abc</p></body.content>
```
Because we assume that the body will _always_ contains newline break, as such, we pass to XML parser two node **without parent**

In addition to fixing above issues, we also add some fixes:
- Handle `h2` tag (Superdesk client only support this header in body), though we saw that most of provided files had its body be wrapped in `p` tag
- Remove linebreak (`<p><br/></p>`) and empty tag (`<p>   </p>` or `<p></p>`)

SDBELGA-263

### DIFF
--- a/server/belga/io/feed_parsers/belga_afp_newsml_1_2.py
+++ b/server/belga/io/feed_parsers/belga_afp_newsml_1_2.py
@@ -8,10 +8,8 @@
 # AUTHORS and LICENSE files distributed with this source code, or
 # at https://www.appendsourcefabric.org/superdesk/license
 
-import re
-
 from superdesk.io.registry import register_feed_parser
-
+from superdesk.text_utils import get_text
 from .base_belga_newsml_1_2 import BaseBelgaNewsMLOneFeedParser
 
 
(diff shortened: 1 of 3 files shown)


server/belga/io/feed_parsers/belga_afp_newsml_1_2.py
```diff
@@ -51,8 +51,17 @@ def parser_newsitem(self, item, newsitem_el):
             })
         # add content for headline when it is empty
         if item.get('urgency') in ('1', '2') and not item.get('headline'):
-            first_line = item.get('body_html', '').strip().split('\n')[0]
-            first_line = etree.fromstring(first_line).text
+            first_line = ''
+            # remove trailing whitespace in p or h2 tag and empty tag
+            body = re.sub(r'(>)(\s*)(.*?)(\s*)(<)', r'\1\3\5', item.get('body_html', ''))
```
petrjasek: it's kind of hard to understand all those regexes, what about something like:
```python
from superdesk.text_utils import get_text

for line in get_text(body).split('\n'):
    if line.strip():
        item['headline'] = 'URGENT: {}'.format(line.strip())
        break
```

idsvn: There is a small problem, get_text will combine all p tag text as a single paragraph, for example, from the ticket attachment files
`
<body.content><p>Un homme poignarde des passants à Villejuif, la police le "neutralise" (préfecture de police)</p><p>sm/jt/bma</p><p/></body.content>
`

so get_text will return
`Un homme poignarde des passants à Villejuif, la police le "neutralise" (préfecture de police)sm/jt/bma
`

but the content includes two paragraphs, so we think the headline should be
`Un homme poignarde des passants à Villejuif, la police le "neutralise" (préfecture de police)
`

petrjasek: you can use this [lf_on_block](https://github.com/superdesk/superdesk-core/blob/develop/superdesk/text_utils.py#L45)

idsvn: Sorry we missed that, with it  `get_text` is working as we expected :+1:

PR #132, opened 2020-02-06
setup profile for video/picture items

petrjasek:
and populate default values using client extension

SDBELGA-280


client/extensions/iptc/src/extension.ts
```diff
@@ -0,0 +1,61 @@
...
+const DEFAULT_SUBJECT = {
+    'media-credit': ['BELGA_ON_THE_SPOT'],
+    'belga-keywords': ['RUSHES', 'BELGAILLUSTRATION', 'BELGAINTERVIEW', 'BELGAINSERT'],
+    'countries': ['BEL'],
+};
+
+const extension: IExtension = {
+    id: 'iptc',
+    activate: (superdesk: ISuperdesk) => {
+        console.info('activate');
```
ride90: forgotten console

client/extensions/iptc/src/extension.ts
```diff
@@ -0,0 +1,61 @@
+import {IExtension, IExtensionActivationResult, ISuperdesk, IArticle, ISubject} from 'superdesk-api';
+
+type Scheme = 'media-credit' | 'belga-keywords' | 'countries';
+
+const CVS: Array<Scheme> = ['media-credit', 'belga-keywords', 'countries'];
+
+const DEFAULT_SUBJECT = {
+    'media-credit': ['BELGA_ON_THE_SPOT'],
+    'belga-keywords': ['RUSHES', 'BELGAILLUSTRATION', 'BELGAINTERVIEW', 'BELGAINSERT'],
+    'countries': ['BEL'],
```
ride90: If you need a single selection, you can use a `country` CV https://github.com/superdesk/superdesk-belga/blob/master/server/data/vocabularies.json#L4204

server/picture-profile.json
```diff
@@ -0,0 +1,196 @@
...
+            "order": 7,
+            "section": "header"
+        },
+        "event_description": {
+            "order": 8,
+            "section": "header",
+            "sdWidth": "full",
+            "textarea": true
+        },
+        "source": {
```
ride90: if `source` is always "Belga" and filled with extension anyway, I would remove it from the editor.

server/data/vocabularies.json
```diff
@@ -11874,6 +11874,25 @@
             }
         ]
     },
+    {
+        "_id": "media-credit",
```
ride90: I think 
```
"service": { "all": 1}
```
is missed.

server/picture-profile.json
petrjasek: I think if it's not there it won't make it to the formatter

ride90: I've tried and it worked. I hope it was not cache :)

petrjasek: ok thx for checking

PR #133, opened 2020-02-06
Belga NewsML updates

ride90:
- [x] SDBELGA-274 [BelgaNewsML] Export slugline as Topic in administrative metadata.

- [x] SDBELGA-279 [BelgaNewsML] Published NewsML should contain all objects in the update chain

- [x] SDBELGA-278 [BelgaNewsML] Author initials should be exported instead of full names
- [x] SDBELGA-282 [BelgaNewsML] Combine metadata of picture/video in 'Description' metadata field

depends on https://github.com/superdesk/superdesk-core/pull/1803


server/belga/publish/belga_newsml_1_2.py
```diff
@@ -436,10 +445,31 @@ def _format_picture(self, newscomponent_1_level, picture):
...
+            if key == 'description_text' \
+                    and picture.get('extra', {}).get('people') \
+                    and picture.get('extra', {}).get('event_description'):
+                # format caption using `people` and `event_description` fields
+                data_content = '{} on the picture regarding {}'.format(
+                    picture['extra']['people'], picture['extra']['event_description']
+                )
+            elif key == 'description_text' \
+                    and picture.get('extra', {}).get('people') \
+                    and not picture.get('extra', {}).get('event_description'):
```
petrjasek: I would say it's not needed here as if that part didn't pass the `if` branch would run

server/belga/publish/belga_newsml_1_2.py
```diff
@@ -587,10 +617,31 @@ def _format_video(self, newscomponent_1_level, video):
...
+            if key == 'description_text' \
+                    and video.get('extra', {}).get('people') \
+                    and video.get('extra', {}).get('event_description'):
+                # format caption using `people` and `event_description` fields
+                data_content = '{} on the video regarding {}'.format(
+                    video['extra']['people'], video['extra']['event_description']
+                )
+            elif key == 'description_text' \
+                    and video.get('extra', {}).get('people') \
+                    and not video.get('extra', {}).get('event_description'):
```
petrjasek: same here this line is always true

server/belga/publish/belga_newsml_1_2.py
```diff
@@ -804,6 +855,11 @@ def _format_administrative_metadata(self, newscomponent_2_level, item):
                 administrative_metadata, 'Property',
                 {'FormalName': 'Priority', 'Value': str(item['priority'])}
             )
+        if 'slugline' in item:
```
petrjasek: imo it's better in general to check for `item.get('slugline')` so it also checks if it's not `None` which might break the output

PR #144, opened 2020-02-19
fix config for media profile

petrjasek:
and handle default values from parent item in extension

SDBELGA-280


client/extensions/iptc/src/extension.ts
```diff
@@ -36,16 +38,38 @@ const extension: IExtension = {
                                     translations: subj.translations,
                                 });
                             });
-                    }
+                    });
 
                     Object.assign(item, {
+                        language: 'nl',
```
tomaskikutis: Don't mutate the item that's being passed as an argument. Instead, you can create another item and mutate that if you need to.

```
const itemNext = {
    ...item,
    language: 'nl',
    // ...
};
```

PR #146, opened 2020-02-20
fix brief internal destination macro

petrjasek:
- ignore items with non TEXT profile and with long body
- fix time selected when de-scheduling the item
- schedule new item 30m after publishing text in case it's scheduled
- handle validation error for newly create items to avoid client error

SDBELGA-315 SDBELGA-316 SDBELGA-318 SDBELGA-319

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: python
 
-python: "3.5"
+python: "3.6"
 
 services:
     - mongodb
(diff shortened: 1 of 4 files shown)


GyanP:
What was the issue with this https://dev.sourcefabric.org/browse/SDBELGA-316?

petrjasek:
this brief macro publishes another item with brief and the errors from publishing that second item were propagated to the response

PR #153, opened 2020-02-26
use docker for services

petrjasek:


ride90:
I think we can also run belga specific tests:

```
jobs:
    include:
        - script: cd server && pytest/nosetests
          name: "belga-pytest/nosetest"
```

petrjasek:
I think it's still there https://github.com/superdesk/superdesk-belga/pull/153/files#diff-354f30a63fb0907d4ad57269548329e3R20

PR #154, opened 2020-02-26
SDBELGA-334 Exported NewsML has empty CreditLine field

ride90:


ride90:
Discussed it with Karel and Stefaan. It'll require some modifications.

ride90:
This PR is a temporary solution. There are a bunch of changes are coming. After the output format is fully tested, I will make refactoring to keep things a little bit logical and simpler.

PR #155, opened 2020-02-28
[SDBELGA-336] Added config for publishing associations

karelpetrak:
# publishing of associated and related items
PUBLISH_ASSOCIATED_ITEMS = True


ride90:
why do we need this?

ride90:
for Belga news ml all associated items + entire update/translate chain is published anyway :) 
I can not imagine what is gonna happened when we enable this setting.

PR #175, opened 2020-03-13
bump version

tomaskikutis:
SDBELGA-292

use develop so custom avatars are available


petrjasek:
we should stick with the release not develop for belga, so if we need some feature there now it should be ported to 1.33

PR #194, opened 2020-04-01
Update search_cvs [SDBELGA-247]

karelpetrak:
Added "sources" and "languages"

### DIFF
--- a/client/superdesk.config.js
+++ b/client/superdesk.config.js
@@ -62,7 +62,9 @@ module.exports = function(grunt) {
         
         search_cvs: [
             {id: 'belga-keywords', name:'Belga keywords', field: 'subject', list: 'belga-keywords'},
-            {id: 'services-products', name:'Packages', field: 'subject', list: 'services-products'}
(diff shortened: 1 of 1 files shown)


client/superdesk.config.js
```diff
@@ -62,7 +62,9 @@ module.exports = function(grunt) {
         
         search_cvs: [
             {id: 'belga-keywords', name:'Belga keywords', field: 'subject', list: 'belga-keywords'},
-            {id: 'services-products', name:'Packages', field: 'subject', list: 'services-products'}
+            {id: 'services-products', name:'Packages', field: 'subject', list: 'services-products'},
+            {id: 'sources', name:'Sources', field: 'subject', list: 'sources'},
+            {id: 'languages', name:'Language', field: 'subject', list: 'languages'}
```
petrjasek, line 67: I think the field should be `source` and `language`

ride90: We have CVs `sources` and `languages`, I think it should work.

PR #199, opened 2020-04-08
Changed color for urgency 5

karelpetrak:
to yellow #FFFF00


petrjasek:
you might need to increase init_version

PR #200, opened 2020-04-09
[SBELGA-271][SDBELGA-288] Belga Archive Advanced Search

idsvn:
1. Display date for Archive search item (SDBELGA-271)
  Map `firstcreated` and `versioncreated` (used in [listItems](https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/apps/search/components/ListItemInfo.tsx#L56) > [firstLine](https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/apps/search/constants.ts#L52))
  Should we also map `firstpublished` field for using in [Belga NewsML export](https://github.com/superdesk/superdesk-belga/blob/master/server/belga/publish/belga_newsml_1_2.py#L333)?
2. Add advanced search, search parameters are provided in **SDBELGA-173**:
- Search by headline is removed, based on @karelpetrak 's feedback, user should use full text search
- Search by **period** will override search by **date**
- For 24h period, because belga APIs only provide search by date, so we will display result for both _today and yesterday_

SDBELGA-271
SDBELGA-288

### DIFF
--- a/server/belga/search_providers.py
+++ b/server/belga/search_providers.py
@@ -289,7 +289,7 @@ class Belga360ArchiveSearchProvider(superdesk.SearchProvider):
             return ''
 
     def _get_period(self, period):
-        today = arrow.now(BELGA_TZ)
+        today = arrow.now(superdesk.app.config['DEFAULT_TIMEZONE'])
         return {
             'fromDate': today.shift(**self.PERIODS.get(period)).format('YYYYMMDD'),
             'toDate': today.format('YYYYMMDD'),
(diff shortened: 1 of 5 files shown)


server/belga/search_providers.py
```diff
@@ -259,6 +282,19 @@ def api_get(self, endpoint, params):
...
         return resp.json()
 
+    def _get_belga_date(self, date):
+        try:
+            return arrow.get(date, 'DD/MM/YYYY').format('YYYYMMDD')
+        except arrow.parser.ParserError:
+            return ''
+
+    def _get_period(self, period):
+        today = arrow.now(BELGA_TZ)
```
ride90, line 292: it's better to use `superdesk.app.config['DEFAULT_TIMEZONE']`

PR #209, opened 2020-04-27
update planning to 1.33rc

petrjasek:


client/package.json
```diff
@@ -7,8 +7,6 @@
     "dependencies": {
         "superdesk-analytics": "github:tomaskikutis/superdesk-analytics#next",
```
MarkLark86, line 8: This can now change to use the official analytics repo

petrjasek: which branch?

MarkLark86: The other repos are using the release branch, so I would suggest the same
`github:superdesk/superdesk-analytics#release/1.33`

PR #215, opened 2020-05-05
Expected behavior for related content of translations

ride90:
SDBELGA-384
SDBELGA-378

- group signals into package
- during translation (manual or macro), belga-360-archive/videos are removed from a translation
- all non-text items inherit the main item’s language (even if non-text items have an explicit language) during publishing, regardless if the main item is a translation or not.

depends on https://github.com/superdesk/superdesk-core/pull/1863


server/belga/signals/__init__.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+def init_app(_app):
+    # generate id for belga url
+    item_create.connect(generate_id_for_url.handle_create)
+    item_update.connect(generate_id_for_url.handle_update)
+    # unmark user when moved to incoming stage
+    item_move.connect(unmark_user_when_moved_to_incoming_stage.unmark_user)
+    # change profile from ALERT to TEXT on update
+    item_rewrite.connect(update.handle_update)
```
petrjasek, line 16: I think you have still some connect calls in those modules

server/belga/signals/handle_translate.py
```diff
@@ -0,0 +1,22 @@
+import logging
+from copy import deepcopy
+
+from superdesk.metadata.item import CONTENT_TYPE
+
+
+logger = logging.getLogger(__name__)
```
petrjasek, line 7: seems like not very useful here

PR #218, opened 2020-05-13
Fix flask import

pablopunk:


petrjasek:
can you also update planning/analytics to use develop branches?

PR #222, opened 2020-05-15
Enable ES mapping for archive.extra

ride90:
SDBELGA-387
⚠️ requires ES reindex

### DIFF
--- a/server/settings.py
+++ b/server/settings.py
@@ -173,7 +173,31 @@ SCHEMA_UPDATE = {
             'schema': {},
             'mapping': {
                 'type': 'object',
-                'enabled': True
+                'properties': {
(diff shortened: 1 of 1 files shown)


server/settings.py
```diff
@@ -166,6 +166,20 @@
...
 }
 
+SCHEMA_UPDATE = {
+    'archive': {
+        'extra': {
+            'type': 'dict',
+            'schema': {},
+            'mapping': {
+                'type': 'object',
+                'enabled': True
```
petrjasek, line 176: it would be good to define the fields explicitly so there won't be any issues with auto type detection

ride90: updated

PR #223, opened 2020-05-18
Join belga_remote_newsml12 and belganewsml12 feed parsers into one

ride90:
SDBELGA-389


server/belga/io/feed_parsers/belga_newsml_1_2.py
```diff
@@ -169,6 +197,10 @@ def parser_newsitem(self, newsitem_el):
                 # NewsComponent
                 try:
                     self.parser_newscomponent(item, news_component_2)
+                    # SDBELGA-328
+                    if item.get('abstract'):
+                        abstract = '<p>' + item['abstract'] + '</p>'
+                        item['body_html'] = abstract + item['body_html']
```
idsvn, line 203: From the ticket, this is Belga Remote specific logic, right?

ride90: I've clarified it with Stefaan and it should be applied to all belga news ml inputs.

PR #228, opened 2020-06-01
[SDBELGA-394] Automatically save article on 'Coming Up' change

idsvn:
@tomaskikutis How could we update `etag` of currently editing article? It seems currently superdesk does not support this?
We think fixing `etag` may help with [this issue on video editor](https://github.com/superdesk/superdesk-client-core/pull/3218#discussion_r355555321)

SDBELGA-394


tomaskikutis:
> How could we update etag of currently editing article?

We should try not to do it at all when possible. For example, in this PR, instead of sending a patch HTTP request, you could add another method to the extensions API which would trigger saving the article. The same code would get executed as clicking the button manually, and there wouldn't be any etag issues. It could work similarly as adding an image [here](https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/core/superdesk-api.d.ts#L965).

By the way, can you move this code to belga repository? https://github.com/superdesk/superdesk-belga/tree/master/client/extensions

idsvn:
Good point, we will start working on it

ride90:
@idsvn , what's the status of this PR. It would be good to include it in a new release.

idsvn:
> @idsvn , what's the status of this PR?
> It would be good to include it in a new release.

It still in progress, it will be done in next week

ride90:
@idsvn any updates on it?

idsvn:
> @idsvn any updates on it?

Waiting for https://github.com/superdesk/superdesk-client-core/pull/3527

client/extensions/saveArticleOnComingUpChange/src/extension.ts
```diff
@@ -0,0 +1,32 @@
...
+                authoring: {
+                    onUpdateAfter: (current: IArticle, next: IArticle) => {
+                        if (
+                            // Only trigger save on toggle
+                            (current.extra?.DueBy == null || next.extra?.DueBy == null) &&
+                            current.extra?.DueBy !== next.extra?.DueBy
+                        ) {
+                            superdesk.ui.article.save();
+                        }
+                        return Promise.resolve(next);
```
tomaskikutis, line 22: `onUpdateAfter` doesn't return anything, so you should remove this line

PR #235, opened 2020-06-19
Use ES7 in fireq

ride90:


petrjasek:
I think it should get it from superdesk-core

PR #242, opened 2020-07-03
[SDBELGA-410] Add profile for belga archive items

idsvn:
Display profile for Belga Archive items
- Use profile in `content_profiles.json` instead of `content_profiles` service as we have discussed
- If profile is `Short`, use `Text`, and leave it empty if it does not exist in superdesk

SDBELGA-410

### DIFF
--- a/server/belga/search_providers.py
+++ b/server/belga/search_providers.py
@@ -235,8 +235,7 @@ class Belga360ArchiveSearchProvider(superdesk.SearchProvider):
         super().__init__(provider)
         self.session = requests.Session()
         self.content_types = {
-            c['_id']: c['_id']
-            for c in superdesk.get_resource_service('content_types').find({})
+            c['_id'] for c in superdesk.get_resource_service('content_types').find({})
         }
 
     def url(self, resource):
@@ -322,8 +321,10 @@ class Belga360ArchiveSearchProvider(superdesk.SearchProvider):
     def _get_profile(self, profile):
         label = profile.lower()
         if label == 'short':
-            return self.content_types.get('text')
-        return self.content_types.get(label)
+            label = 'text'
+        if label not in self.content_types:
+            return
+        return label
 
     def format_list_item(self, data):
         guid = '%s%d' % (self.GUID_PREFIX, data['newsObjectId'])
--- a/server/belga/search_providers.py
+++ b/server/belga/search_providers.py
@@ -1,6 +1,7 @@
 import hmac
 import uuid
 import arrow
+import json
 import hashlib
 import requests
 import superdesk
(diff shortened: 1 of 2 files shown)


server/belga/search_providers.py
```diff
@@ -234,6 +235,8 @@ class Belga360ArchiveSearchProvider(superdesk.SearchProvider):
     def __init__(self, provider):
         super().__init__(provider)
         self.session = requests.Session()
+        with open('data/content_types.json') as f:
```
ride90, line 238: can we just read this info from db?

idsvn: We use `content_types.json` to avoid mapping the wrong profile if there is two profile with the same label.
Belga instance exists some duplicate label due to new content profiles added to the json file, normally label should be unique. So user need to:
- Enable only one profile if there is multiple profile with the same label
- OR manually delete duplicate profile

What do you think?

ride90: @idsvn , you can use `_id` from the DB, it's unique:

```
>>> db.getCollection('content_types').find({}, {_id: 1})

{
    "_id" : "alert"
}

/* 2 */
{
    "_id" : "brief"
}

/* 3 */
{
    "_id" : "tip"
}

/* 4 */
{
    "_id" : "text"
}

/* 5 */
{
    "_id" : ObjectId("5ecfa296b0be54a1722ae9cb")
}

/* 6 */
{
    "_id" : ObjectId("5efb4a9caec6c0fbdaf98905")
}
```

server/belga/search_providers.py
```diff
@@ -234,6 +234,10 @@ class Belga360ArchiveSearchProvider(superdesk.SearchProvider):
     def __init__(self, provider):
         super().__init__(provider)
         self.session = requests.Session()
+        self.content_types = {
+            c['label'].lower(): c['_id']
```
ride90, line 238: I think we should use `_id` as a key because Id is unique.

ride90: see this comment https://github.com/superdesk/superdesk-belga/pull/242#discussion_r454880864

server/belga/search_providers.py
```diff
@@ -234,6 +234,10 @@ class Belga360ArchiveSearchProvider(superdesk.SearchProvider):
     def __init__(self, provider):
         super().__init__(provider)
         self.session = requests.Session()
+        self.content_types = {
+            c['_id']: c['_id']
```
idsvn, line 238: We do look up more often (`format_list_item` for each item) so `dict` is more effective here than `set`

ride90: @idsvn why do you think `dict` is more effective? I think in our case it does not matter actually because we don't have millions of items there, so having `set` is logical.

PR #252, opened 2020-09-18
[SDBELGA-253] Add search by date for Belga Image

idsvn:
- Remove options search by `Last * hours`
- Add search by Date
- Upgrade `superdesk-analytics` to `v1.33.4`, this does not affect server, as we are using `develop` branch

*For Belga Image API, please check SDBELGA-9*

SDBELGA-253

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -5,7 +5,7 @@
         "postinstall": "node ./tasks/install-extensions.js"
     },
     "dependencies": {
-        "superdesk-analytics": "github:superdesk/superdesk-analytics#release/1.33",
+        "superdesk-analytics": "github:superdesk/superdesk-analytics#v1.33.4",
         "superdesk-core": "github:superdesk/superdesk-client-core#develop",
         "superdesk-planning": "github:superdesk/superdesk-planning#develop"
     }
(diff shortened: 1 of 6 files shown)


client/package.json
```diff
@@ -5,7 +5,7 @@
         "postinstall": "node ./tasks/install-extensions.js"
     },
     "dependencies": {
-        "superdesk-analytics": "github:superdesk/superdesk-analytics#release/1.33",
+        "superdesk-analytics": "github:superdesk/superdesk-analytics#v1.33.4",
```
ride90, line 8: We should also change server side analytics version to keep it in sync with client one

ride90 (changes requested):

client/package.json
idsvn: Server is using `develop` branch so we think it's okay to keep it, should we pin it to `v1.33.4` as well?

https://github.com/superdesk/superdesk-belga/blob/0b07b29b816f1693907e409818a706f55225324b/server/requirements.txt#L8

ride90: IMO it's better to have it in sync, to have the same version on client and on server.
please, try to fix a build.

idsvn: I checked and it seems the build is failed on Node v8 but no problem for Node v10 and v12. Could you check which version of Nodejs travis is using?

ride90 (approved):
@idsvn , let's merge it and after I will update travis config and after we can fix an issue with CI

PR #253, opened 2020-10-01
Add move attachment config for BelgaNewsMLOneFeedParser

ride90:
and update travis config, add mypy and fix mypy errors


server/belga/io/feed_parsers/belga_newsml_1_2.py
```diff
@@ -55,6 +55,7 @@ class BelgaNewsMLOneFeedParser(BaseBelgaNewsMLOneFeedParser):
         }
     }
     SUPPORTED_BINARY_ASSET_SUBTYPES = ('SOUND', 'CLIP', 'COMPONENT', 'IMAGE')
+    MOVE_FILE = False
```
ride90, line 58: this part might be changed to True, depends on the Stefaan's reply

petrjasek (approved):

PR #260, opened 2020-11-04
Recognition of BelgaRemote metadata

ride90:
SDBELGA-428

### DIFF
--- a/server/belga/io/feed_parsers/belga_newsml_1_2.py
+++ b/server/belga/io/feed_parsers/belga_newsml_1_2.py
@@ -718,7 +718,36 @@ class BelgaNewsMLOneFeedParser(BaseBelgaNewsMLOneFeedParser):
             for element in elements:
                 # country
                 if element.attrib.get('FormalName', '') == 'Country' and element.attrib.get('Value'):
-                    item.setdefault('extra', {})['country'] = element.attrib.get('Value')
+                    country_name_regex = {
+                        '$regex': r'^{}$'.format(element.attrib.get('Value')),
+                        # case-insensitive
+                        '$options': 'i'
+                    }
+                    scheme = 'countries'
+                    country = get_resource_service('vocabularies').get_from_mongo(
+                        req=None,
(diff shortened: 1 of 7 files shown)


server/belga/io/feed_parsers/belga_newsml_1_2.py
```diff
@@ -718,7 +723,36 @@ def parse_descriptivemetadata(self, item, descript_el):
...
+                        '$regex': r'^{}$'.format(element.attrib.get('Value')),
+                        # case-insensitive
+                        '$options': 'i'
+                    }
+                    scheme = 'countries'
+                    country = get_resource_service('vocabularies').get_from_mongo(
+                        req=None,
+                        lookup={
+                            '_id': 'countries',
+                            'items.translations.name.{}'.format(item['language']): country_name_regex
```
petrjasek, line 736: seems like duplicate to me given that you do the projection with the same filter

petrjasek (approved):

server/belga/io/feed_parsers/belga_newsml_1_2.py
ride90: It's not a duplicate, with lookup I query the entire `countries` CV with all items, but with the filter I get [only one](https://github.com/superdesk/superdesk-belga/pull/260/files#diff-1a85f6b207213178d2925e41a08bfa36bfe5c3a22acbfb72dab9d6a3ebea3f2eR749) specified `item`. I can do this with python, but I think doing it on the mongo level is better.

petrjasek: yep with the lookup you will only find the countries cv if it has that country_name, then you filter the items, without it you would always find the countries cv and then filter the items, so it could be empty at the end

ride90: ahh, I've got your point now. When I was doing it I was thinking that no sense to run a filter if lookup is already empty.

PR #261, opened 2020-12-14
[SDBELGA-420] BelgaPress search

idsvn:
BelgaPress API has not support filtering by `country`, `sources` and `sourcetype` yet, we will make another PR when it is available.
Supported filter:
- Period
- Content type
- Language

SDBELGA-420


ride90 (approved):

server/belga/search_providers.py
```diff
@@ -361,7 +362,136 @@ def format_list_item(self, data):
...
+            'client_id': self.provider.get('config', {}).get('username'),
+            'client_secret': self.provider.get('config', {}).get('password'),
+            'grant_type': 'client_credentials'
+        })
+        resp.raise_for_status()
+        if resp.status_code == 200 and resp.content:
+            data = resp.json()
+            self._access_token = data.get('access_token')
+
+    def find(self, query: Dict[str, Any], params: Optional[Dict] = None):
```
idsvn, line 402: It would be nice to have a typing module (`superdesk.typing` or `superdesk.typing.search_provider`) which declare [precise type](https://mypy.readthedocs.io/en/stable/more_types.html#typeddict) for common dict object like `query`, `params`

PR #262, opened 2020-12-30
Sort out content profiles naming in staging and production

ride90:
SDBELGA-436

We store fixed content profiles in the json file and these profiles were imported during init data cmnd.
On Belga stage/prod instance content profiles were also created manually (which is wrong), so at the end, it was a conflict between imported (correct) profiles and manually created. 
This PR does next steps:
- replace **manually created** `data.profile` id with **the content profile from json** in `content_templates`
- replace **manually created** `default_content_profile` id with **the content profile from json** in `desks`
- replace **manually created** `profile` id with **the content profile from json** in `archive`
- replace **manually created** `profile` id with **the content profile from json** in `archive.associations.*.profile`
- replace **manually created** `profile` id with **the content profile from json** in `published.associations.*.profile`
- replace **manually created** `profile` id with **the content profile from json** in `archive_autosave.associations.*.profile`
- replace **manually created** `profile` id with **the content profile from json** in `archive_versions.associations.*.profile`
- remove manually created content types

P.S. I've tested it with DB dump from the stage instance.

### DIFF
--- a/server/belga/macros/brief_internal_routing.py
+++ b/server/belga/macros/brief_internal_routing.py
@@ -1,9 +1,8 @@
 
 import logging
 import superdesk
-from typing import List, Dict
+from typing import List
 
-from flask import current_app as app
 from datetime import timedelta
 from superdesk.metadata.item import CONTENT_STATE, PUBLISH_SCHEDULE, SCHEDULE_SETTINGS
 from superdesk.macros.internal_destination_auto_publish import internal_destination_auto_publish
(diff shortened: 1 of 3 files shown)


petrjasek:
you only update content items in mongo, so are you assuming there will be re-index from mongo done after?

ride90:
@petrjasek , yes, on a stage instance there are around 150k items so it should not take to much time. Prod instance has even less.

petrjasek (approved):

PR #269, opened 2021-01-28
Switch from an ALERT to a TEXT does not always modify the Distribution and Priority fields. [SDBELGA-426]

GyanP:


client/extensions/updateArticleOnProfileChange/src/extension.ts
```diff
@@ -19,9 +19,9 @@ const extension: IExtension = {
...
                             superdesk.entities.contentProfile.get(next.profile),
                         ]).then((res) => {
-                            const [currentProfile, nextProfile] = res;
+                            const nextProfile = res[1];
 
-                            if (currentProfile.label === 'ALERT' && nextProfile.label !== 'ALERT') {
+                            if (nextProfile.label !== 'ALERT') {
                                 return Promise.resolve({
                                     ...next,
                                     urgency: 3,
```
petrjasek, line 27: imo that would always reset the urgency, or could you change it?

GyanP: Good catch.

PR #273, opened 2021-02-01
Ingest KeywordLine as belga-keywords.

ride90:
SDBELGA-428
related to https://github.com/superdesk/superdesk-core/pull/2132

### DIFF
--- a/server/belga/io/feed_parsers/belga_newsml_1_2.py
+++ b/server/belga/io/feed_parsers/belga_newsml_1_2.py
@@ -608,32 +608,15 @@ class BelgaNewsMLOneFeedParser(BaseBelgaNewsMLOneFeedParser):
         # belga-keywords
         for element in newslines_el.findall('KeywordLine'):
             if element is not None and element.text:
-                scheme = 'belga-keywords'
-                belga_keyword_name_regex = {
-                    '$regex': r'^{}$'.format(element.text.strip()),
-                    # case-insensitive
-                    '$options': 'i'
-                }
-                cursor = get_resource_service('vocabularies').get_from_mongo(
-                    req=None,
-                    lookup={'_id': scheme},
-                    projection={
-                        'items': {
-                            '$elemMatch': {
-                                'translations.name.{}'.format(item['language']): belga_keyword_name_regex
-                            }
-                        }
-                    }
+                belga_keywords = get_resource_service("vocabularies").get_items(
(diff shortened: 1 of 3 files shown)


server/belga/io/feed_parsers/belga_newsml_1_2.py
```diff
@@ -605,10 +605,36 @@ def parse_newslines(self, item, newslines_el):
...
         for element in newslines_el.findall('KeywordLine'):
             if element is not None and element.text:
-                item.setdefault('keywords', []).append(element.text)
+                scheme = 'belga-keywords'
+                belga_keyword_name_regex = {
+                    '$regex': r'^{}$'.format(element.text.strip()),
+                    # case-insensitive
+                    '$options': 'i'
+                }
+                cursor = get_resource_service('vocabularies').get_from_mongo(
```
petrjasek, line 617: not sure how many keywords they usually have, but would it be better to fetch them all once and do the lookup via python?

petrjasek (approved):

server/belga/io/feed_parsers/belga_newsml_1_2.py
ride90: I thought about it, they usually have 1-4 keywords (1-4 queries per article). 
There are a lot of items in `belga-keywords` CV, looping through all items every time might be less effective than one query (not sure actually 😄 ) .

I can make something like that:
```
belga_keywords_map = {
    'fr': {
        ...
        'Liège': {
            "name" : "LIEGE",
            "qcode" : "LIEGE",
            "is_active" : true,
            "translations" : {
                "name" : {
                    "nl" : "Luik",
                    "fr" : "Liège"
                }
            }
        }
        ...
    },
    'nl': {
        ...
        'Luik': {
            "name" : "LIEGE",
            "qcode" : "LIEGE",
            "is_active" : true,
            "translations" : {
                "name" : {
                    "nl" : "Luik",
                    "fr" : ""
                }
            }
        }
        ...
    }
}
```

not really readable but the search will be instant.

what do you think?

petrjasek: actually I don't mind the queries, but would be good to have some method on the service to find values by name and have the implementation there, it seems like it's a bit low level for the parser

petrjasek (approved):

PR #284, opened 2021-03-09
Modify 'Set Default Metadata With Translate' macro to keep Storytags (keywords). [SDBELGA-458]

GyanP:


MarkLark86 (commented):
It would be ideal to have tests for these to make sure they work now and in the future

GyanP:
> It would be ideal to have tests for these to make sure they work now and in the future

sure.

MarkLark86 (approved):

PR #309, opened 2021-04-26
Sdbelga 500 internal attachments are output to belga  newsml 1.2

dev-tsoni:


petrjasek:
it's hard to read all those changes, would something like:
```py
if attachment.get('internal'):
    return
```
work too? that would be easier to read imo and would avoid all that indentation

dev-tsoni:
> it's hard to read all those changes, would something like:
> 
> ```python
> if attachment.get('internal'):
>     return
> ```
> 
> work too? that would be easier to read imo and would avoid all that indentation

PR #316, opened 2021-05-07
Change the correspondent author role to editor on translation [SDBELGA-517]

GyanP:


server/belga/macros/update_translation_metadata_macro.py
```diff
@@ -0,0 +1,57 @@
...
+            # update author role correspondent to editor
+            correspondent.update(
+                {"name": editor["name"], "role": editor["qcode"], "_id": editor_id}
+            )
+
+    return item
+
+
+name = "Update Translation Metadata Macro"
+label = lazy_gettext("Update Translation Metadata Macro")
```
petrjasek, line 54: could we use some more descriptive name maybe? and how is this hooked into translations? shouldn't it set this from language etc?

petrjasek (approved):

PR #318, opened 2021-05-13
FIX: Some published stories are marked for users and cannot be unmarked [SDBELGA-512]

GyanP:


server/data_updates/00030_20210512-155559_archive.py
```diff
@@ -0,0 +1,35 @@
...
+
+    Refer to https://dev.sourcefabric.org/browse/SDBELGA-512 for more information
+    """
+
+    resource = 'archive' # will use multiple resources, keeping this here so validation passes
+
+    def forwards(self, mongodb_collection, mongodb_database):
+        for resource in ["archive", "published"]:
+            service = get_resource_service(resource)
+            for doc in service.find({"marked_for_user": {"$exists": True}, "state": "published"}):
```
petrjasek, line 26: not sure if just published state is the problem, what about updated/corrected/unpublished et?

GyanP: means the stories whose state is "in_progress" and "submitted" will be marked for user and all others like kill/corrected/published etc should be unmarked, right?

petrjasek: I think that none of published stories should be marked, but any idea how that could have happened?

GyanP: I'm quite not sure why this happened because we already have that removed mark for the user functionality for the published story here https://github.com/superdesk/superdesk-core/blob/develop/apps/publish/content/publish.py#L63. Yes, you are right the mark for the user should be removed for all published stories.

petrjasek (approved):

PR #324, opened 2021-05-25
implement multiple belga coverages in single item

petrjasek:
SDBELGA-519 SDBELGA-452

### DIFF
--- a/client/extensions/belgaCoverage/src/belga-coverage-carousel.tsx
+++ b/client/extensions/belgaCoverage/src/belga-coverage-carousel.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react';
 
 import {ISuperdesk} from 'superdesk-api';
-import {Carousel, IconButton} from 'superdesk-ui-framework/app-typescript';
+import {Carousel, IconButton} from 'superdesk-ui-framework';
 import {IBelgaCoverage, getCoverageInfo, getCoverageImages, IBelgaImage} from './belga-image-api';
 import 'superdesk-ui-framework/dist/superdesk-ui.bundle.css';
 
@@ -66,7 +66,7 @@ export default class BelgaCoverageCarousel extends React.PureComponent<IProps, I
         const {preview} = this.props;
         const {coverage, images} = this.state;
 
-        const headerActions = preview !== true ? (
+        const headerMeta = preview !== true ? (
             <React.Fragment>
                 <time>{formatDate(coverage.createDate)}</time>
                 {this.props.removeCoverage != null && (
@@ -88,7 +88,7 @@ export default class BelgaCoverageCarousel extends React.PureComponent<IProps, I
                 title={coverage.name}
                 imageCount={coverage.nrImages}
                 description={coverage.description}
-                headerActions={headerActions}
+                headerMeta={headerMeta}
             />
         );
     }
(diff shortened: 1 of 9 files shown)


client/extensions/belgaCoverage/src/belga-coverage-preview.tsx
```diff
@@ -2,28 +2,28 @@ import * as React from 'react';
...
-                        {props.item.type !== 'graphic' &&
-                            <BelgaCoverageInfo coverageId={id} superdesk={superdesk} />
-                        }
-                        <BelgaCoverageImages
-                            coverageId={id}
-                            rendition={'preview'}
-                            maxImages={10}
-                            superdesk={superdesk}
-                        />
+                        {props.item.type !== 'graphic' && (
```
tomaskikutis, line 16: Is it really supposed to work for all types except graphic?

If so, it'd be nicer to use a ternary

```
{
    props.item.type === 'graphic'
        ? (
            <Component1 />
        )
        : (
            <Component2 />
        )
}
```

client/extensions/belgaCoverage/src/belga-coverage-carousel.tsx
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    render() {
+        if (this.state.loading) {
+            return (
+                <div className="sd-loader" />
+            );
+        }
+
+        const {Alert} = this.props.superdesk.components;
```
tomaskikutis, line 57: If you are targeting `develop` branch of the client, it's now possible to import `superdesk` in [the root scope of the file](https://github.com/tomaskikutis/superdesk-planning/blob/assignments-widget-in-the-toolbar/client/planning-extension/src/assignments-overview/assignments-overview-list-item.tsx#L7) now

client/extensions/belgaCoverage/src/belga-coverage-carousel.tsx
```diff
@@ -0,0 +1,112 @@
...
+                <div className="sd-thumb-carousel__item-inner">
+                    <img src={props.thumbnailUrl} alt={props.name} />
+                </div>
+            </div>
+        );
+
+        const headerTemplate = (
+            <div className="sd-thumb-carousel__header">
+                <h4 className="sd-thumb-carousel__heading">{coverage.name}</h4>
+                <Badge text={'' + coverage.nrImages} type='light' />
```
tomaskikutis, line 81: Is `nrImages` image count? It'd be good to rename it or at least add a comment in the interface if the name comes from their API.

client/extensions/belgaCoverage/src/belga-coverage-preview.tsx
```diff
@@ -2,28 +2,27 @@ import * as React from 'react';
...
-                        }
-                        <BelgaCoverageImages
-                            coverageId={id}
-                            rendition={'preview'}
-                            maxImages={10}
-                            superdesk={superdesk}
-                        />
+                        {props.item.type !== 'graphic' ? ( // belga coverage in an article
+                            <BelgaCoverageCarousel coverageId={id} superdesk={superdesk} preview={true} />
+                        ) : ( // single belga coverage in the search, there is description already
```
tomaskikutis, line 18: What kind of description are you referring to here?

> belga coverage in the search

do you mean preview pane opened when in search view?

To me it seems like both components render lists of images, why are different components used?

tomaskikutis (approved):

client/extensions/belgaCoverage/src/belga-coverage-preview.tsx
petrjasek: we have search for those coverages, in that case in the preview you open just single coverage item which has name as headline and description as abstract and we only need to show images in preview on top of it, while in the other case it's embedded in an article which doesn't have the coverage name/description in the other metadata, so we display it all together

PR #329, opened 2021-07-06
BRIEF keyword missing in translation text [SDBELGA-492]

GyanP:


server/belga/macros/set_default_metadata_with_translate.py
```diff
@@ -18,6 +18,41 @@
 logger = logging.getLogger(__name__)
 
 
+def set_belga_keywords(item):
+    """This field is used to set belga keyword field with the value "Brief" upon translation
+    """
+    subjects = item.get("subject", [])
```
MarkLark86, line 24: When `subject` is not defined or `None`, does this function update `item["subject"]`, since the return of the `get` function will be a default value here. So when updating this value, the reference may not be to the `item["subject"]` directly but to a locally defined value.

Maybe using `setdefault` would suffice here

MarkLark86:
It would be good to add unit tests under `server/tests/macros/set_default_metadata_with_translate_test.py` to guarantee this functionality will continue to work in the future

MarkLark86 (approved):

PR #335, opened 2021-08-05
As an editor I like to see and search based on all keywords coming with the ingest sources AFP, ANP and dpa [SDBELGA-555]

GyanP:


server/data/vocabularies.json
```diff
@@ -6655,6 +6655,22 @@
         },
         "unique_field": "qcode"
     },
+    {
+		"_id": "original-metadata",
+		"display_name": "original metadata",
+		"type": "manageable",
```
MarkLark86, line 6661: Should this type be `unmanageable` considering it is empty (as the values are retrieved from the item itself)?

MarkLark86 (approved):

PR #336, opened 2021-08-13
Test instace for SDBELGA-555

GyanP:


petrjasek:
do we still need this @GyanP ?

PR #338, opened 2021-08-16
remove package lock

tomaskikutis:


tomaskikutis:
@petrjasek why does CI fail without lock file, but doesn't in other repos? https://github.com/superdesk/superdesk-ntb/pull/418

It'd be good not to have lock files in develop branch so it's quicker to update

petrjasek:
that's due to using `npm ci` in https://github.com/superdesk/superdesk-belga/blob/develop/.github/workflows/tests.yml#L82
you can change that to `npm ci || npm install` that will work with or without lock

petrjasek (approved):

PR #341, opened 2021-09-02
move startingDay option from js file to database

tomaskikutis:
SDESK-5991


petrjasek:
imo there are more non-default configs eg timezone

tomaskikutis:
I'm only moving 2 configs to UI at the moment, the others remain in `superdesk.config.js`. The other one is `minutesOnline`, but since no customer has it set, I don't need an upgrade script.

After I merge that big PR on client core, I'll work on moving other configs.

petrjasek (approved):

PR #347, opened 2021-09-20
Fix: Recognition of 'Country', 'Belga Keywords' and 'Country Keywords of AFP, ANP, DPA. [SDBELGA-521]

GyanP:
similar to SDBELGA-521, we have to store data for ANP and DPA also.

### DIFF
--- a/server/belga/io/feed_parsers/belga_dpa_newsml_2_0.py
+++ b/server/belga/io/feed_parsers/belga_dpa_newsml_2_0.py
@@ -21,6 +21,7 @@ from superdesk.errors import ParserError
 from superdesk.metadata.item import CONTENT_TYPE
 
 from .belga_newsml_mixin import BelgaNewsMLMixin
+from superdesk import get_resource_service
 
 logger = logging.getLogger(__name__)
 NS = {'xhtml': 'http://www.w3.org/1999/xhtml',
--- a/server/belga/io/feed_parsers/belga_dpa_newsml_2_0.py
+++ b/server/belga/io/feed_parsers/belga_dpa_newsml_2_0.py
@@ -175,14 +175,18 @@ class BelgaDPANewsMLTwoFeedParser(BelgaNewsMLMixin, NewsMLTwoFeedParser):
             if name_elt is not None:
                 item.setdefault('extra', {})['city'] = name_elt.text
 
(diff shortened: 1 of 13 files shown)


petrjasek:
is there any other ticket related to it @GyanP ?

GyanP:
> is there any other ticket related to it @GyanP ?

No, for AFP we already created the PR, and its merged as well and the ticket is closed but as per the comment mentioned in the ticket, the client wants the same changes for ANP and DPA also so created this one.

server/belga/io/feed_parsers/belga_dpa_newsml_2_0.py
```diff
@@ -175,14 +176,18 @@ def parse_content_meta(self, tree, item):
...
+            belga_keywords = get_resource_service('vocabularies').get_items(_id='belga-keywords', qcode=data.upper())
+            if len(belga_keywords) > 0:
+                item.setdefault('subject', []).append(belga_keywords[0])
+            else:
+                # save all other unmapped keywords as subject with scheme original-metadata
+                item.setdefault('subject', []).append({
+                    'name': data,
+                    'qcode': data,
+                    'scheme': 'original-metadata'
+                })
```
petrjasek, line 190: seems like this part is more or less duplicatedin base parser, can you avoid that duplication?

server/belga/io/feed_parsers/belga_newsml_mixin.py
```diff
@@ -34,3 +34,17 @@ def _get_countries(self, country_code):
             for c in countries
             if c.get('qcode') == country_code.lower() and c.get('is_active')
         ]
+
+    def _get_keywords(self, data):
+        belga_keywords = get_resource_service('vocabularies').find_one(req=None, _id='belga-keywords').get('items', [])
+
+        _belga_keyword_list = [
+            {'name': c['name'], 'qcode': c['qcode'], 'translations': c['translations'], 'scheme': 'belga-keywords'}
```
petrjasek, line 42: why not use the `get_items` method?

GyanP: my bad, I will update it with get_items.

petrjasek (approved):

PR #355, opened 2021-10-05
As an editor I like to see and search based on all keywords coming with the ingest sources EFE, Itartass [SDBELGA-566]

GyanP:
store all the possible keywords for EFE and itartass


petrjasek:
it seems like this would only populate it for keywords which don't match belga keywords, but is that actually according to the specs? it looks to me like there should be all of them no matter if they match belga keywords or not

petrjasek:
ok, it would probably not allow same qcode anyhow with 2 schemas

petrjasek (approved):

PR #356, opened 2021-10-08
As an editor I would like to delay the automated publishing of certain news items during night time [SDBELGA-588]

GyanP:


server/belga/macros/brief_internal_routing.py
```diff
@@ -135,6 +135,12 @@ def brief_internal_routing(item: dict, **kwargs):
     update_schedule_settings(item, PUBLISH_SCHEDULE, item[PUBLISH_SCHEDULE])
     item[PUBLISH_SCHEDULE] = item[PUBLISH_SCHEDULE].replace(tzinfo=None)
 
+    # Do not auto publish between 4 to 7 am
+    publish_time = utc_to_local(item[SCHEDULE_SETTINGS]['time_zone'], published_at).time()
```
petrjasek, line 139: we should rather use `app.config["DEFAULT_TIMEZONE"]` as that item time zone could be different

server/belga/macros/brief_internal_routing.py
```diff
@@ -135,6 +135,12 @@ def brief_internal_routing(item: dict, **kwargs):
...
+    if time(4, 00) <= publish_time and time(7, 00) >= publish_time:
+        logger.info('macro stopped:do not autopublish between 4 to 7 am item=%s', guid)
+        raise StopDuplication()
```
petrjasek, line 142: would it actually create the item if we do this?

petrjasek (approved):

PR #364, opened 2021-10-27
test instance for SDBELGA-588: Set item publish schedule to 16:30 h for autopublish between 15 to 16 h

GyanP:


GyanP:
Only for testing

PR #366, opened 2021-11-11
As a user I want to have the Coming Up field active when I create a new story, add current time T + 45 minutes [SDBELGA-466]

GyanP:


petrjasek:
I'm not sure if we should hardcode it here if we have it already configured via UI

GyanP:
> I'm not sure if we should hardcode it here if we have it already configured via UI

actually, we have to set while creating new article current time + 45 mins that we have to set it there I think because in UI we will have the enable/disable the toggle for coming upfield in the template.

petrjasek:
yep in the template we set the toggle on, and then when we create the item I guess it should read the value from the field config, so imo it should be all handled via core without additional code

GyanP:
Okay, I have done the changes in the core so closing this one

GyanP:
Revert old changes which are not required now, as we did changes in core https://github.com/superdesk/superdesk-core/pull/2298

petrjasek (approved):

PR #367, opened 2021-11-23
Ampersand character incorrectly handled in archive search result [SDBELGA-591]

GyanP:


server/belga/search_providers.py
```diff
@@ -26,6 +26,8 @@ def get_text(value, strip_html=True):
     try:
         text = value.strip()
         if strip_html:
+            # Preserve ampersand character
+            text = text.replace("&", "&amp;")
```
petrjasek, line 30: I wonder if we shouldn't rather use some html.escape, but anyhow would be good to test this

GyanP: Yep, I will add a test case for it.

server/tests/search_providers/belga_360_archive_test.py
```diff
@@ -102,7 +102,7 @@ def test_format_list_item(self):
         ))
         self.assertEqual(item['body_html'], (
             '&nbsp;&nbsp;&nbsp;&nbsp;'
-            'Morbi lacus ex, molestie id ullamcorper quis, scelerisque quis lectus.<br/>&nbsp;&nbsp;&nbsp;&nbsp;'
+            'Morbi lacus ex, molestie id ullamcorper quis & scelerisque quis lectus.<br/>&nbsp;&nbsp;&nbsp;&nbsp;'
```
petrjasek, line 105: shouldn't it be `&amp;` here?

GyanP: Nope, `&amp;` parse into `&` so here we will get `&`.

petrjasek: so without that there would be no `&` ? I wonder how it would fail

GyanP: yep, imo in parse_html here https://github.com/superdesk/superdesk-core/blob/develop/superdesk/etree.py#L113 after when we convert it to string then somehow it missed.

petrjasek: I see, but still shouldn't it be rather escaped like the other entities?

GyanP: imo I think it is better as we have to preserve ampersand because like in the string we have `&` and internally we are converting `&` into `&amp;` then in response we are getting again `&`, so simply passing ampersand **&** and getting in response **&**.

petrjasek: ok, so can you maybe just test it with that co&v to be sure it works there?

GyanP: yep, I have tested, but if you say so in the test case instead of testing just **&** i will test with **co&v**.

petrjasek (approved):

PR #371, opened 2021-12-03
FIX: exception is thrown that prevents added translation fields in metadata from being saved [SDBELGA-601]

GyanP:


petrjasek (commented):
also bump init_version so it actually gets updated next time

petrjasek (approved):

PR #372, opened 2021-12-13
test SDESK-6323

GyanP:
#testing instance


petrjasek:
@GyanP do we need this one still?

PR #374, opened 2021-12-22
Elements missing in the Belga Archive [SDBELGA-470]

GyanP:


server/belga/search_providers.py
```diff
@@ -332,7 +335,7 @@ def _get_body_html(self, item):
         return body.replace('\n', '<br/>&nbsp;&nbsp;&nbsp;&nbsp;')
 
     def _get_abstract(self, item):
-        return self._get_newscomponent(item, 'lead')
+        return '&nbsp;&nbsp;&nbsp;&nbsp;' + get_text(self._get_newscomponent(item, 'lead')) + '<br/><br/>'
```
petrjasek, line 338: why those spaces in front?

server/belga/search_providers.py
```diff
@@ -369,14 +373,34 @@ def format_list_item(self, data):
             'creditline': get_text(data['credit']),
             'source': get_text(data['source']),
             'language': get_text(data['language']),
-            'abstract': get_text(self._get_abstract(data)),
-            'body_html': self._get_body_html(data),
+            'body_html': self._get_abstract(data) + self._get_body_html(data),
```
petrjasek, line 376: would be better probably to do the join here than adding `br` to abstract so it's obvious how those are connected

server/belga/search_providers.py
```diff
@@ -369,14 +373,34 @@ def format_list_item(self, data):
...
             '_fetchable': False,
+            'keywords': data.get("keywords"),
+            'sign_off': self.get_sign_off(data.get('authors')),
+            'subject': self.get_subjects(data.get('keywords'))
         }
 
+    def get_sign_off(self, authors):
+        if not authors:
+            return
+        return ', '.join(map(lambda x: x['name'] + '/' + x['type'], authors))
```
petrjasek, line 389: imo it should be only names joined by `/`

petrjasek (commented):
pls test also sign_off and keywords

server/belga/search_providers.py
GyanP: so that it looks similar to the way we have body_html on UI.

server/belga/search_providers.py
GyanP: According to a comment mentioned by Philippe on the ticket, the sign-off should be like `PHD/Admin`, here **Admin** is the status of the author.

server/belga/search_providers.py
GyanP: I have `br` tags for `'&nbsp;&nbsp;&nbsp;&nbsp;' + get_text(self._get_newscomponent(item, 'lead')) + '<br/><br/>'` to separate abstract and body html and also body html comes with new line
for example :

            """Abstract content """
                                          -------->one Blank line
            """Body HTML content"""

petrjasek: yep what I mean is it would be more readable like:
```
'body_html': self._get_abstract(data) + '<br/><br/>' + self._get_body_html(data),
```
so you can see those are connected via blank line, which is not really part of the abstract

server/belga/search_providers.py
petrjasek: admin is author role?

server/belga/search_providers.py
GyanP: yep got it.

server/belga/search_providers.py
GyanP: Yep, on the preview of sign_off, it will show like `name/type` so here it is like `authors : [{'name': 'PHD', 'type': 'ADMIN'}]`

petrjasek: it's usually sign off of multiple users, so more like `author1/author2/author3` than `author1/role1`

GyanP: supppse we have mulitple author for example `authors : [{'name': 'PHD', 'type': 'ADMIN'}, {'name': 'BRV', 'type': 'AUTHOR'}]`  so its sign off  will be like `PHD/Admin, BRV/Author`

server/belga/search_providers.py
petrjasek: and we do it when we create content? or when we ingest something? I thought that's just css

server/belga/search_providers.py
petrjasek: yep but if you would do the same thing via superdesk it would be `PHD/BRV` so not sure what is the expectation

server/belga/search_providers.py
GyanP: here you can see they have added the spaces in front of body HTML https://github.com/superdesk/superdesk-belga/blob/ed02e586f1a36c0bf0044e5c14c95d1d2298d800/server/belga/search_providers.py#L334 so I have spaces for abstract also so that both looks good and the same.

server/belga/search_providers.py
```diff
@@ -332,7 +335,7 @@ def _get_body_html(self, item):
         return body.replace('\n', '<br/>&nbsp;&nbsp;&nbsp;&nbsp;')
 
     def _get_abstract(self, item):
-        return self._get_newscomponent(item, 'lead')
+        return '&nbsp;&nbsp;&nbsp;&nbsp;' + get_text(self._get_newscomponent(item, 'lead'))
```
petrjasek, line 338: I still don't think we need those spaces before the very first paragraph

server/belga/search_providers.py
petrjasek: I see, but then when we use such story with different sign off format how does it work in the output?

GyanP: With superdesk we have sign-off `PHD/BRV`
With Belga archive will have sign off `PHD/Admin, BRV/Author` (as per the request of Phillippe here we always get the response in list format `[{'name': 'PHD', 'type': 'ADMIN'}, {'name': 'BRV', 'type': 'AUTHOR'}]` )

GyanP: > With superdesk we have sign-off `PHD/BRV` With Belga archive will have sign off `PHD/Admin, BRV/Author` (as per the request of Phillippe here we always get the response in list format `[{'name': 'PHD', 'type': 'ADMIN'}, {'name': 'BRV', 'type': 'AUTHOR'}]` )
do you know any other format for Belga archive sign-off?

petrjasek: I'm talking about output format, if we get such article from belga archive and publish it what will be there? does it match the output of article created via superdesk?

petrjasek (approved):

PR #375, opened 2022-01-05
Picture of translation is not exported to NewsML on publication [SDBELGA-597]

GyanP:


server/belga/publish/belga_newsml_1_2.py
```diff
@@ -418,10 +418,11 @@ def _format_picture(self, newscomponent_1_level, picture):
         :param Element newscomponent_1_level: NewsComponent of 1st level
         :param dict picture: picture item
         """
-        guid = picture.get('guid') or picture['_id']
-        if guid in self._seen_pictures:
+        # picture with same language not exported multiple times in output
+        image_id = (picture.get('guid') or picture['_id']) + picture.get('language', 'en')
```
petrjasek, line 422: does that language play any role? the language wouldn't change imo unless the guid/id changes

GyanP: Yes, Before what happening here is, this check is only for `guid/id` so after translation of the main item it is creating only one picture item, 
but now we have a check that `guid/id + language` so after translation we have two picture items in the output XML

For example: suppose we have one main item with one picture association(current language NL) then while translating the main item(suppose we converted from NL to FR language) and then publish the main item-
**Before:** 

> In output XML we have two main item(one for NL and another for FR) and only one picture item(for NL only)(according to SDBELGA-514)

**Now:** 

> In output XML we have two main items (one for NL and another for FR) and two picture items (for NL and FR also)(according to SDBELGA-597)

**I made changes as per the comment mentioned by Gideon and that is** 

> Here is the explanation from the original ticket in which we excluded the image from the payload **#SDBELGA-514**
> You have closed ticket **#SDBELGA-597** because the request conflicts with this ticket, where we had asked you not to export the same image ID twice.
> 
> It appears that this request was incorrect. Instead of deduplicating based on the image ID only, it should be based on the combination (image ID + language), which is the unique combination to our business. Therefore I reopen this ticket and ask you to alter the deduplication rule accordingly.

petrjasek: that all looks about right, but the picture language will be the same, when you translate an article we don't translate it's media, and if we did then that would create new picture items with different id, so this wouldn't be a problem in the first place. the language there should be the text item language.

server/tests/publish/belga_newsml_1_2_items_chain_tests.py
```diff
@@ -680,6 +680,8 @@ def test_belga_text_newscomponent(self):
             ('urn:belga.be:coverage:6690595', 'nl'),
             ('urn:belga.be:360archive:77777777', 'nl'),
             ('original-fr', 'fr'),
+            ('urn:belga.be:image:154620545', 'fr'),
+            ('urn:belga.be:image:154670415', 'fr'),
```
petrjasek, line 684: imo here the language of those images should be nl, that's not changing

GyanP: here we already have for NL https://github.com/superdesk/superdesk-belga/blob/develop/server/tests/publish/belga_newsml_1_2_items_chain_tests.py#L678-#L679, I think we are changing for NL to FR so we need fr also right?

petrjasek: when we translate text with pictures we don't translate the media inside, do we?

GyanP: I just revert those changes of SDBELGA-514 because as per the new changes and before SDBELGA-514 also it is working the same. 
and as far as I know, we are not translating pictures (not sure) but after translation when we publish the main item then in the output XML we have two picture items as I have shared in the previous comment(https://github.com/superdesk/superdesk-belga/pull/375#discussion_r780089322) so we need those old changes here.

server/belga/publish/belga_newsml_1_2.py
petrjasek: seems like the language is inherited from the text item, can you add a comment regarding this?

petrjasek (approved):

petrjasek (approved):

PR #381, opened 2022-02-22
FIX: Author of related article is overwritten by author of main news [SDBELGA-620]

GyanP:


petrjasek (commented):
how does it fix the issue? can you add a test maybe?

petrjasek (approved):

PR #382, opened 2022-02-25
Import Belga biographies into Superdesk [SDBELGA-623]

GyanP:
Belga is not using the Abstract field, we already mapped the abstract data into the body_html field.

### DIFF
--- a/server/belga/io/feed_parsers/belga_newsml_1_2.py
+++ b/server/belga/io/feed_parsers/belga_newsml_1_2.py
@@ -11,6 +11,7 @@
 import os
 import hashlib
 import logging
+import itertools
 from io import BytesIO
 from copy import deepcopy
 from uuid import uuid4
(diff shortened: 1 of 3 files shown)


server/belga/io/feed_parsers/belga_newsml_1_2.py
```diff
@@ -739,19 +741,20 @@ def parse_descriptivemetadata(self, item, descript_el):
...
+                    if countries:
+                        item.setdefault('subject', []).extend(countries + self._get_country(countries[0]["qcode"]))
+
                 # city
                 if element.attrib.get('FormalName', '') == 'City' and element.attrib.get('Value'):
                     item.setdefault('extra', {})['city'] = element.attrib.get('Value')
 
+        # remove duplicated subject
+        item['subject'] = [
+            dict(i) for i, _ in itertools.groupby(sorted(item['subject'], key=lambda k: k['qcode']))
```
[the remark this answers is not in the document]
GyanP: I have checked it and found that it will work perfectly fine. here we are just sorting using qcode then groupby will automatically remove duplicate subjects.

server/tests/io/feed_parsers/belga_newsml_1_2_test.py
```diff
@@ -212,18 +222,6 @@ def test_content(self):
         self.assertEqual(item["source"], "ANP/BELGA")
         self.assertEqual(item["extra"], {'city': 'ANTWERPEN'})
         self.assertEqual(item["type"], "text")
-        self.assertEqual(
```
petrjasek, line 215: imo it would be good to check it's actually not there, eg. `self.assertNotIn('abstract', item)`

GyanP: can we do this `self.assertEqual(item['abstract'], '')` as we will have a key in the item but it will be always empty?

petrjasek: yep that's ok

petrjasek (approved):

PR #384, opened 2022-03-17
Improve the Belga archive integration [SDBELGA-606]

GyanP:


GyanP:
1. #SDBELGA-633 related changes.
2. It contains the solution of:
- content types should allow being multiselection to combine various content types in one search
- Related items associated with a text (or other content types) should be referenced in the preview of the main item

server/belga/search_providers.py
```diff
@@ -297,9 +300,30 @@ def find(self, query, params=None):
...
+        docs = [self.get_detailed_info(item) for item in data[self.items_field]]
         return BelgaListCursor(docs, data[self.count_field])
 
+    def get_detailed_info(self, item):
+        try:
+            detailed_resp = self.api_get('archivenewsitems/' + str(item['newsItemId']), {})
+            formatted_data = self.format_list_item(detailed_resp)
+            formatted_data["associations"] = self.get_related_article(detailed_resp)
+            return formatted_data
+        except Exception as e:
```
petrjasek, line 312: I think we should catch 404 here and not just any error

server/belga/search_providers.py
```diff
@@ -297,9 +300,30 @@ def find(self, query, params=None):
             api_params['searchText'] = ''
 
         data = self.api_get(self.search_endpoint, api_params)
-        docs = [self.format_list_item(item) for item in data[self.items_field]]
+        docs = [self.get_detailed_info(item) for item in data[self.items_field]]
```
petrjasek, line 303: I'm not sure if we can do this performance wise, doing extra request for each item means quite some load. imo would be better to only fetch the detail for item preview, that might require some core changes

server/belga/search_providers.py
```diff
@@ -274,7 +277,7 @@ def find(self, query, params=None):
 
             api_params['assetType'] = ' OR '.join([
                 _type for _type in self.TYPE_SUPPORT
```
petrjasek, line 279: I think it might be good to have some defaults without related items

server/belga/search_providers.py
GyanP: yep, I was thinking the same, we have to do changes in core probably so I will check it.

server/belga/search_providers.py
GyanP: Do you mean if in params we do not have any `type` then we just have to set some default? Currently, defaults are all content types if `type` is not in params -
>    **'Text', 'Brief', 'Alert', 'Short', 'Coverage', 'RelatedDocument', 'RelatedArticle', 'Audio', 'Picture', 'SMS', 'Video'**

should we set default only below four types?

> **'Text'**, **'Brief'**, **'Alert'**, **'Short'**

petrjasek: yep, but we should probably rather make those selected by default in the UI so it's more obvious

GyanP: means all the content types(Text, Brief, RelatedArticle etc) should be selected by default in UI when we open Belga Archive search provider right @petrjasek?

petrjasek: right

client/belga/360archive/BelgaSearchPanelController.ts
```diff
@@ -1,27 +1,49 @@
...
+  $scope: any;
+  $rootScope: any;
+  static $inject: string[];
+  constructor($scope, $location, $rootScope) {
+    this.$scope = $scope;
+    this.$location = $location;
+    this.$rootScope = $rootScope;
+    let defaultSelected = {
+      "Alert": true, "Text": true, "Brief": true, "Short": true, "Coverage": true,
+      "RelatedDocument": true, "RelatedArticle": true, "Audio": true, "Picture": true, "SMS": true, "Video": true
```
petrjasek, line 22: I would make the second line all `false` if that means unselected by default

petrjasek (approved):

PR #386, opened 2022-03-21
Import Belga biographies into Superdesk [SDBELGA-623]

GyanP:


GyanP:
Authors are not previewing on UI for ingested biographies XML articles, and also getting console errors of duplicate authors so to fix it.

petrjasek (approved):

PR #391, opened 2022-05-18
check the Country mapping for AFP ingests [SDBELGA-636]

GyanP:


petrjasek (commented):
seems like it would parse buch of metadata which we don't care about,
the tickets seems to be related to City only.
and test pls

GyanP:
> seems like it would parse buch of metadata which we don't care about, the tickets seems to be related to City only. and test pls

Sounds right, I have address the comments.

server/belga/io/feed_parsers/belga_afp_newsml_1_2.py
```diff
@@ -21,7 +21,7 @@ class BelgaAFPNewsMLOneFeedParser(BaseBelgaNewsMLOneFeedParser):
     """Feed Parser for Belga specific AFP NewsML."""
 
     NAME = "belga_afp_newsml12"
-    label = "Belga specific AFP News ML 1.2 Parser"
+    lpel = "Belga specific AFP News ML 1.2 Parser"
```
petrjasek, line 24: typo

petrjasek (approved):

PR #392, opened 2022-06-22
Fix Subjects in belga AFP newsml 1.2 feedparser

GyanP:
SDBELGA-653


petrjasek:
what ticket is that?

GyanP:
> what ticket is that?

https://sofab.atlassian.net/browse/SDBELGA-653

petrjasek (approved):

PR #394, opened 2022-07-15
ingest and distribute ANSA news agency [SDBELGA-661]

devketanpro:


server/tests/io/feed_parsers/belga_ansa_test.py
```diff
@@ -0,0 +1,99 @@
...
+                },
+            ],
+        )
+        self.assertEqual(item["source"], "ANSA")
+        self.assertEqual(item["anpa_category"], [{"qcode": "ALR", "name": "ALR"}])
+
+        expected_body = (
+            '<lang lang="it"/>\n'
+            "            <block>\n                \n                "
+            '<img source="05be579082cc8e4649d31ccd14aec976.jpg"/>\n                '
```
petrjasek, line 93: I think we should start with `p` and skip the `lang` and `block` and `img`

server/tests/io/feed_parsers/belga_ansa_test.py
```diff
@@ -0,0 +1,99 @@
...
+                "located": {
+                    "city_code": "ROME",
+                    "city": "ROME",
+                    "tz": "UTC",
+                    "dateline": "city",
+                }
+            },
+        )
+        self.assertEqual(
+            item["headline"], ">>>ANSA/Gazprom cuts Italy gas supplies by a third"
```
petrjasek, line 63: I think we won't need that `>>>ANSA/` prefix

server/tests/io/feed_parsers/belga_ansa_test.py
```diff
@@ -0,0 +1,99 @@
...
+                "parent": "NEWS",
+                "scheme": "services-products",
+            },
+        ]
+        expected_subjects.sort(key=lambda i: i["scheme"])
+        self.assertEqual(item["subject"], expected_subjects)
+        self.assertEqual(item["guid"], "XAM22192010243_AMZ_X083")
+        self.assertEqual(item["priority"], 5)
+        self.assertEqual(
+            item["keywords"], [">>>ANSA/Gazprom cut its flow of gas into Italy"]
```
petrjasek, line 44: I would remove the prefix `>>>ANSA/` here

server/belga/io/feed_parsers/belga_ansa.py
```diff
@@ -115,13 +117,15 @@ def meta_parse(self, xml, item):
         item["source"] = (source_elem.text).strip()
 
         # headline
-        item["headline"] = self.get_headline(xml).strip()
+        headline_text = self.get_headline(xml).strip()
+        if headline_text.startswith(">>>ANSA/"):
+            item["headline"] = headline_text[len(">>>ANSA/") :]
```
petrjasek, line 122: would be good to have a function for this to avoid duplication

petrjasek (approved):

server/belga/io/feed_parsers/belga_ansa.py
devketanpro: I'll probably go with it since it seems like a nice idea.

PR #395, opened 2022-08-04
ingest and distribute ANSA news agency [SDBELGA-661]

devketanpro:


petrjasek (approved):

server/belga/io/feed_parsers/belga_ansa.py
```diff
@@ -216,6 +216,7 @@ def parse_content(self, xml):
     def remove_prefix(self, text):
```
petrjasek, line 216: would be good to have type annotation or something so this would actually not pass the CI

PR #397, opened 2022-09-13
Populate the News ML creditline field based on the "source name" of the ingest provider [SDBELGA-672]

devketanpro:


server/belga/publish/belga_newsml_1_2.py
```diff
@@ -868,7 +868,17 @@ def _format_newslines(self, newscomponent_2_level, item):
...
-        SubElement(newslines, 'CreditLine').text = self.DEFAULT_CREDITLINE
+
+        #SDBELGA-672
+        for subject in item.get('subject', []):
+            if subject.get('scheme') == 'sources':
+                SubElement(newslines, 'CreditLine').text = subject.get("name")
+
+        if len(newslines.xpath('CreditLine')) == 0:
+            if item.get('source'):
+                SubElement(newslines, 'CreditLine').text = item['source'].strip("incoming")
```
petrjasek, line 879: it should handle both `incoming` and `Incoming`, they are not consistent there. and also remove the space between name and incoming pls

server/belga/publish/belga_newsml_1_2.py
```diff
@@ -864,55 +953,73 @@ def _format_newslines(self, newscomponent_2_level, item):
...
+        SubElement(newslines, "DateLine")
+        SubElement(newslines, "HeadLine").text = item.get("headline")
+        SubElement(newslines, "CopyrightLine").text = item.get("copyrightholder")
+
+        # SDBELGA-672
+        for subject in item.get("subject", []):
+            if subject.get("scheme") == "sources":
+                SubElement(newslines, "CreditLine").text = subject.get("name")
+
+        if len(newslines.xpath("CreditLine")) == 0:
```
petrjasek, line 966: imo this might be a bit more readable if it would be like:
```python
for subject in item.get("subject", []):
    if subject.get("scheme") == "sources":
        SubElement(newslines, "CreditLine").text = subject.get("name")
        break
else:
    if item.get("source"):
        ...
```
so it's more obvious those blocks belong together

petrjasek (approved):

PR #399, opened 2022-09-21
As a Belga admin I want to ingest audio news based on the Belga News ML file format into Superdesk [SDBELGA-676]

devketanpro:


server/tests/io/feed_parsers/belga_newsml_1_2_test.py
```diff
@@ -392,3 +392,36 @@ def test_audio_ingest(self):
...
+                return BytesIO(f.read())
+
+        parser._get_file = MagicMock(side_effect=get_file_side_effect)
+
+        with open(fixture, 'rb') as f:
+            self.xml_root = etree.parse(f).getroot()
+            self.item = parser.parse(self.xml_root, provider)
+
+    def test_audio_ingest(self):
+
```
petrjasek, line 419: can you pls assert len on `self.item` length so it's obvious how many items are parsed from the file?

devketanpro: Sure

server/belga/io/feed_parsers/belga_newsml_1_2.py
```diff
@@ -45,7 +45,7 @@ class BelgaNewsMLOneFeedParser(BaseBelgaNewsMLOneFeedParser):
 
     label = 'Belga News ML 1.2 Parser'
 
-    SUPPORTED_TEXT_ASSET_TYPES = ('ALERT', 'SHORT', 'TEXT', 'BRIEF', 'ORIGINAL')
+    SUPPORTED_TEXT_ASSET_TYPES = ('ALERT', 'SHORT', 'TEXT', 'BRIEF', 'ORIGINAL', 'QUOTE')
     SUPPORTED_MEDIA_ASSET_TYPES = {
```
petrjasek, line 49: I would expect the QUOTE rather in the AUDIO section of the SUPPORTED_MEDIA_ASSET_TYPES, that wouldn't work?

devketanpro: I've tried before, but it didn't work.

petrjasek: ok but we should rather fix it, it's confusing to have audio type together with those text ones

devketanpro: So you mean we have to remove **audio** from the `SUPPORTED_MEDIA_ASSET_TYPES` ?

petrjasek: I mean we should remove QUOTE from SUPPORTED_TEXT_ASSET_TYPES and have it in the AUDIO section,
because it's not text we're ingesting there

server/belga/io/feed_parsers/belga_newsml_1_2.py
```diff
@@ -526,7 +517,7 @@ def parse_newscomponent_media(self, item, newscomponent_el):
                 ))
                 raise SkipItemException
             role_name = role_name.upper()
-            item[ITEM_TYPE] = getattr(CONTENT_TYPE, role_name)
+            item[ITEM_TYPE] = "QUOTE" if role_name == "QUOTE" else getattr(CONTENT_TYPE, role_name)
```
petrjasek, line 520: the type should be rather `audio`, can you pls add that to the test case?

petrjasek (approved):

PR #403, opened 2022-11-10
Media from archive search is not rendered [SDBELGA-677]

devketanpro:


server/belga/search_providers.py
```diff
@@ -481,7 +481,7 @@ def format_list_item(self, data):
 
         return {
             "type": self.get_type(data.get("assetType", "text")),
-            "mimetype": "application/superdesk.item.text",
+            "mimetype": f"application/superdesk.item.{data.get('assetType','text').lower()}",
```
petrjasek, line 484: would be better to use that `get_type` helper so we're only using types we can handle

server/belga/search_providers.py
```diff
@@ -509,10 +509,27 @@ def format_list_item(self, data):
             "sign_off": self.get_sign_off(data.get("authors")),
             "authors": self.get_authors(data.get("authors")),
             "subject": self.get_subjects(data),
+            "renditions": self.get_renditions(data)
+            if data["assetType"] == "Picture"
```
petrjasek, line 513: not sure if the `assetType` is always set, given that we're using `data.get('assetType')` before

server/tests/search_providers/belga_360_archive_test.py
```diff
@@ -326,3 +326,32 @@ def get_period(period):
...
         self.assertEqual(get_period("year"), "20190214")
+
+    def test_get_image_renditions(self):
+        item = self.provider.format_list_item(get_belga360_item()[3])
+        guid = "urn:belga.be:360archive:46768825"
+        self.assertEqual(item["_id"], guid)
+        self.assertEqual(item["mimetype"], "application/superdesk.item.picture")
+        self.assertEqual(item["state"], "published")
+        self.assertEqual(item["guid"], guid)
+        self.assertEqual(item["headline"], "FILES - FBL - WC - 2022")
```
petrjasek, line 337: pls also add a test for description

server/belga/search_providers.py
devketanpro: here we check the `assetType` which we get from the api response, and I checked several items we got `assetType` in every item.

petrjasek: yep but when detecting the type we use it with `.get`, so either we don't need it there or we should use `.get` here as well to be consistent

server/tests/search_providers/belga_360_archive_test.py
```diff
@@ -326,3 +326,33 @@ def get_period(period):
...
+        self.assertEqual(item["description_text"], "")
```
petrjasek, line 338: it shouldn't be empty, but rather this one https://github.com/superdesk/superdesk-belga/pull/403/files#diff-d1bf0aaef3bf81b35878ca4902166f4efe8c7697f94834a98ca59cfe4a91ea7bR312

petrjasek (approved):
:+1:

PR #405, opened 2022-11-21
apply a change to the output formatter in the way the "role" field in the Belga NewsML files is populated [SDBELGA-691]

devketanpro:


petrjasek:
imo it shouldn't be applied when [the map](https://github.com/superdesk/superdesk-belga/blob/develop/server/belga/publish/belga_newsml_1_2.py#L106) is used, only when dealing with content profile

MarkLark86 (approved):

PR #409, opened 2022-11-25
apply a change to the output formatter in the way the role field in the Belga NewsML files is populated [SDBELGA-691]

devketanpro:


petrjasek (commented):
can you add/update a test for this? not sure what does this fix

devketanpro:
> can you add/update a test for this? not sure what does this fix

Okay I will

petrjasek (approved):

PR #411, opened 2022-11-25
Fix : ANP parser creates one string for several metadata values in the field "original metadata" [SDBELGA-689]

devketanpro:


server/belga/io/feed_parsers/belga_anp_newsml_1_2.py
```diff
@@ -70,7 +71,7 @@ def parse_newsitem(self, item, newsitem_el):
             ):
                 subject_list = [
                     {"name": i, "qcode": i, "scheme": "original-metadata"}
-                    for i in set(subject["name"].split(";"))
+                    for i in set(re.split('[- ;]', subject["name"]))
```
petrjasek, line 74: are you sure about the space? how would they send multi word keywords?

devketanpro: we can add pipe symbol `re.split('[-|;]'` for better surety.

petrjasek (approved):

PR #412, opened 2022-11-27
Fix : ANSA news appear with a timestamp one hour ahead of time [SDBELGA-688]

devketanpro:


server/belga/io/feed_parsers/belga_ansa.py
```diff
@@ -221,7 +221,9 @@ def remove_prefix(self, text: str) -> str:
 
     def get_norm_datetime(self, tree):
         value = super().get_norm_datetime(tree)
-        return value.replace(tzinfo=pytz.timezone("CET"))
+        return value.replace(tzinfo=pytz.timezone("CET")).astimezone(
```
petrjasek, line 224: actually this would only work in winter, we need to treat this as a Rome timezone which CET in winter and CEST in summer

petrjasek: here we handle the same kind of input https://github.com/superdesk/superdesk-ansa/blob/develop/server/ansa/parser/parser.py#L181-L183

petrjasek (changes requested):

server/belga/io/feed_parsers/belga_ansa.py
devketanpro: okay, got your point will make changes.Thank you

petrjasek (approved):

PR #413, opened 2022-11-30
Fix the mapping of the City field in the Kyodo parser [SDBELGA-693]

devketanpro:


server/tests/io/fixtures/kyodo_newsml_1_2_belga.xml
```diff
@@ -54,7 +54,6 @@
 				</SubjectCode>
 				<Location HowPresent="Origin">
 					<Property FormalName="Country" Value="FRA"/>
-					<Property FormalName="City" Value="Paris"/>
```
petrjasek, line 57: better keep it there, I suppose they won't remove it from their files

server/belga/io/feed_parsers/belga_kyodo_newsml_1_2.py
```diff
@@ -21,5 +21,20 @@ class BelgaKyodoNewsMLOneFeedParser(BaseBelgaNewsMLOneFeedParser):
...
+    # SDBELGA - 693
+    def parse(self, xml, provider=None):
+        items = super().parse(xml, provider)
+        location_el = xml.find(
+            "NewsItem/NewsComponent/ContentItem/DataContent/nitf/body/body.head/dateline/location"
+        )
+        if location_el is not None:
+            for item in items:
+                if item.get("extra"):
+                    item["extra"]["city"] = location_el.text
```
petrjasek, line 33: you can use `item.setdefault("extra", {})["city"] = location_el.text` instead of that `if else`

PR #419, opened 2022-12-14
As a Belga editor I want to see the search terms highlighted for the Belga Archive search result list and preview [SDBELGA-667]

devketanpro:


server/belga/search_providers.py
```diff
@@ -379,12 +379,12 @@ def find(self, query, params=None):
 
         # SDBELGA-667
         if search_text := api_params.get("searchText"):
-            search_text = " ".join([f"({text})" for text in search_text.split()])
+            search_text = "|".join([f"{text}" for text in search_text.split()])
```
petrjasek, line 382: imo now that `[]` part is only doing the split, so it's like:
```
search_text = "|".join(search_text.split())
```

server/tests/search_providers/belga_360_archive_test.py
```diff
@@ -378,3 +378,31 @@ def test_get_image_renditions(self):
...
+
+    def test_get_highlighted_text(self):
+        with HTTMock(archive_mock):
+            query = {
+                "size": 50,
+                "from": 50,
+                "query": {
+                    "filtered": {
+                        "query": {
+                            "query_string": {"query": "Lorem ipsum"},
```
petrjasek, line 390: can you pls add a test case where it will be the other way around

devketanpro: Yep sure

server/belga/search_providers.py
```diff
@@ -375,6 +376,26 @@ def find(self, query, params=None):
...
+                    highlighted_value = re.subn(
+                        f"({search_text})",
+                        lambda text: " ".join(
+                            [
+                                f'<span class="es-highlight">{s}</span>'
+                                for s in text.groups()
+                            ]
+                        ),
+                        doc.get(field),
+                    )
```
petrjasek, line 395: would probably also use `flags=re.IGNORECASE` so it will match both cases, looks like the search is not case sensitive

petrjasek (approved):

PR #420, opened 2023-01-16
Apply specific packages for agency texts fetched to economic desks [SDBELGA-671]

devketanpro:


petrjasek (changes requested):
according to the ticket it should still assign metadata based on default template, there can be only 1 macro so it needs to be handled here as well

server/belga/macros/add_specific_package.py
```diff
@@ -1,12 +1,16 @@
...
 logger = logging.getLogger(__name__)
 
 
 def update_package(item, **kwargs):
+
+    item = set_default_metadata(item, **kwargs)
+
     if item.get("state") == CONTENT_STATE.FETCHED and item.get("language") in (
         "fr",
         "de",
```
petrjasek, line 16: it says `Dutch` in the ticket that's `nl`

petrjasek (approved):

PR #423, opened 2023-02-09
As a Belga editor I want to see the search terms highlighted for the Belga Archive search result list and preview [SDBELGA-667]

devketanpro:


server/belga/search_providers.py
```diff
@@ -331,6 +327,33 @@ def __init__(self, provider):
...
 
+    def get_search_text(self, query):
+        try:
+            searchText = query["query"]["filtered"]["query"]["query_string"]["query"]
+        except KeyError:
+            searchText = ""
+
+        return searchText
+
+    def get_highlight(self, search_text, docs):
```
petrjasek: should be probably called rather `set_highlight` as it's setting it and not returning anything

petrjasek (approved):

server/belga/search_providers.py
```diff
@@ -352,7 +352,8 @@ def set_highlight(self, search_text, docs):
                     flags=re.IGNORECASE,
                 )
                 if highlighted_value[1]:
-                    doc["es_highlight"] = {field: [highlighted_value[0]]}
+                    doc["es_highlight"] = doc.get("es_highlight", {})
+                    doc["es_highlight"][field] = [highlighted_value[0]]
```
petrjasek: you can do something like:
```python
doc.setdefault("es_highlight", {})[field] = [highlighted_value[0]]
```

petrjasek (approved):

PR #427, opened 2023-03-14
When we perform a fetch of a text from another agency, we want to automatically add the Belga keyword BRIEF in the created text [SDBELGA-715]

devketanpro:


server/belga/macros/set_default_metadata.py
```diff
@@ -92,6 +93,21 @@ def set_default_metadata(item, **kwargs):
...
+    brief_already_exists = any(
+        subject
+        for subject in subject
+        if subject["qcode"] == "BRIEF" and subject["scheme"] == "belga-keywords"
+    )
+    if not brief_already_exists and item.get("state") == CONTENT_STATE.INGESTED:
+        subject.append(
+            {
+                "name": "BRIEF",
+                "qcode": "BRIEF",
```
petrjasek, line 106: this should have also `scheme` defined

petrjasek (approved):

PR #436, opened 2023-08-28
Create only one translation for texts published in double package [SDBELGA-538]

devketanpro:
https://github.com/superdesk/superdesk-core/pull/2467

### DIFF
--- a/server/belga/macros/set_default_metadata_with_translate.py
+++ b/server/belga/macros/set_default_metadata_with_translate.py
@@ -79,28 +79,29 @@ def set_default_metadata_with_translate(item, **kwargs):
         return
 
     # SDBELGA-538
-    filtered_subjects = [
-        s for s in item["subject"] if s.get("scheme") == "services-products"
-    ]
-    internal_destination = kwargs.get("internal_destination", {})
-    content_filter_id = internal_destination.get("filter", "")
+    if item.get("subject"):
+        filtered_subjects = [
+            s for s in item["subject"] if s.get("scheme") == "services-products"
+        ]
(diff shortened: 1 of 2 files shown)


petrjasek (commented):
I woudn't do the filter testing manually here, better use does_match method of content filters service

server/belga/macros/set_default_metadata_with_translate.py
```diff
@@ -79,29 +79,22 @@ def set_default_metadata_with_translate(item, **kwargs):
...
+    ]
+    internal_destination = kwargs.get("internal_destination", {})
+    content_filter_id = internal_destination.get("filter", "")
+    content_filter_service = get_resource_service("content_filters")
+    content_filter = content_filter_service.find_one(req=None, _id=content_filter_id)
+
+    if (
+        len(filtered_subjects) > 1
+        and content_filter
+        and content_filter_service.does_match(content_filter, item)
```
petrjasek: you have to stop it for the item where it didn't match on the first package, so you need to get an item with only first package and then check when it didn't match:
```python3
test_item = item.copy()
test_item["subject"] = [s for s in item["subject"] if s.get("scheme") == "services-products"][:1]  # only first if there is one
...
not content_filter_service.does_match(content_filter, test_item)  # no match on first, this is a duplicate
```

devketanpro: I tried this condition but this will not stop it, because `content_filter_service.does_match(content_filter, test_item)` returns **true**

petrjasek (approved):

PR #437, opened 2023-09-26
Fix : Problem with recognizing of Belga Keywords with ANP texts [SDBELGA-713]

devketanpro:


server/belga/io/feed_parsers/belga_anp_newsml_1_2.py
```diff
@@ -69,12 +70,16 @@ def parse_newsitem(self, item, newsitem_el):
                 subject["scheme"] == "original-metadata"
                 and subject["name"].find(";") != -1
             ):
-                subject_list = [
-                    {"name": i, "qcode": i, "scheme": "original-metadata"}
-                    for i in set(re.split('[-;]', subject["name"]))
-                ]
-                item.setdefault("subject", []).extend(subject_list)
+                for i in set(re.split("[-;]", subject["name"])):
```
petrjasek: might be better to use something else instead of `i`, maybe `keyword`

petrjasek (approved):

PR #438, opened 2023-09-29
City is missing in metadata field while editing an ANP-text [SDBELGA-530]

devketanpro:


server/belga/io/feed_parsers/belga_anp_newsml_1_2.py
```diff
@@ -80,7 +80,21 @@ def parse_newsitem(self, item, newsitem_el):
...
+        if not item.get("extra", {}).get("city"):
+            item.get("extra")["city"] = self.extract_city(item)
+
         return item
 
+    def extract_city(self, item):
+        """
+        extract city from body_html
+        """
+        location_match = re.search(r"([A-Z]+)\s*\(ANP\)", item["body_html"])
```
petrjasek: I suppose there might be cities with more than just 1 word

petrjasek (approved):

PR #440, opened 2024-02-13
fix extensions build with recent client core

petrjasek:


client/extensions/belgaCoverage/src/belga-coverage-editor.tsx
```diff
@@ -1,12 +1,12 @@
...
-import {IEditorComponentProps, ISuperdesk} from 'superdesk-api';
+import {ICommonFieldConfig, IEditorComponentProps, ISuperdesk} from 'superdesk-api';
 
 const ALLOWED = 'application/vnd.belga.coverage';
 const SEPARATOR = ';';
 
-function isAllowedType(event: DragEvent) {
-    return !!event.dataTransfer && event.dataTransfer.types.includes(ALLOWED);
+function isAllowedType(event: DragEvent | undefined) {
+    return event != null && !!event.dataTransfer && event.dataTransfer.types.includes(ALLOWED);
```
thecalcc: might be better to use optional chaining here, so the line would look something like - `event?.dataTransfer?.types?.includes(ALLOWED)`

thecalcc (approved):

PR #441, opened 2024-03-22
related_items field implementation

thecalcc:
SDBELGA-768


thecalcc:
Editor view with some items:
image

Editor view, item hovered:
image


Editor view, no items:
image

Modal view, first item selected, second item hovered:
image

Modal view, item open in preview:
image

MarkLark86:
@thecalcc I don't really see anything here specific to Belga in this PR (except for checking of the search provider with a static string/url).

IMO, this is a core feature and we should do our best to keep it a core feature. In my [Planning PR](https://github.com/superdesk/superdesk-planning/pull/1938) I have added an [improvement](https://github.com/superdesk/superdesk-planning/pull/1938/commits/97d794f542bba8ab9a57f1eabba602d09de5b930) where if you add `EVENT_RELATED_ITEM_SEARCH_PROVIDER_NAME` (with the name of a search provider) to your `settings.py`, this will enable the `related_items` field and allow you to use that config in the front-end (under the `appConfig.planning.event_related_item_search_provider_name` config). Without this config, that field would not be available in the front-end anywhere.

client/extensions/relatedArticles/src/EventsRelatedArticles/EventsRelatedArticlesModal.tsx
```diff
@@ -0,0 +1,307 @@
...
+    Button,
+    WithPagination,
+    Loader,
+    Panel,
+    PanelHeader,
+    PanelContent,
+    PanelContentBlock,
+    ContentDivider,
+} from 'superdesk-ui-framework/react';
+import {getProjectedFieldsArticle, gettext} from 'superdesk-core/scripts/core/utils';
```
MarkLark86, line 16: Please do not import anything from `superdesk-core`, it was something that we used to do years ago before the SuperdeskAPI. Now that we have that API it is best to use that, if needing anything from core then we should add the required functionality to that API.

client/extensions/relatedArticles/src/EventsRelatedArticles/EventsRelatedArticlesModal.tsx
```diff
@@ -0,0 +1,307 @@
...
+        ) {
+            this.setState({
+                loading: true,
+            })
+        }
+    }
+
+    render(): React.ReactNode {
+        const {closeModal} = this.props;
+        const allLanguages = [
```
MarkLark86, line 85: This should ideally come from the languages defined in the system, see [vocabs](https://github.com/superdesk/superdesk-planning/blob/develop/client/selectors/vocabs.ts#L18) selector for an example

client/extensions/relatedArticles/src/EventsRelatedArticles/EventsRelatedArticlesModal.tsx
```diff
@@ -0,0 +1,307 @@
...
+                            style="filled"
+                            type="primary"
+                        />
+                    </Spacer>
+                )}
+            >
+                <Spacer v gap="0">
+                    <div
+                        style={{
+                            padding: 12,
```
MarkLark86, line 142: Constant numbers should not be used, instead we have helper classes for this (https://ui-framework.superdesk.org/#/components/spacing)

client/extensions/relatedArticles/src/EventsRelatedArticles/EventsRelatedArticlesModal.tsx
```diff
@@ -0,0 +1,307 @@
...
+                            backgroundColor: 'white'
```
MarkLark86, line 143: Does this work with a dark theme? An issue with using inline styles vs styles defined in the UI-Framework

client/extensions/relatedArticles/src/EventsRelatedArticles/EventsRelatedArticlesModal.tsx
```diff
@@ -0,0 +1,307 @@
...
+                                    },
+                                ]}
+                            >
+                                {this.state.activeLanguage.label}
+                            </Dropdown>
+                        </SearchBar>
+                    </div>
+                    <ContentDivider margin='none' />
+                    <Spacer h gap="0" justifyContent='start' alignItems='start' noWrap>
+                        <div style={{...{width: this.state.previewItem ? '65%' : '100%'}, ...{padding: 12}}}>
```
MarkLark86, line 176: Surely the width here can be done better, rather than constant percentage values (@fritzSF)?
Again with the padding, please use the helper classes

client/extensions/relatedArticles/src/EventsRelatedArticles/EventsRelatedArticlesModal.tsx
```diff
@@ -0,0 +1,307 @@
...
+                                    display: 'flex',
+                                    width: '100%',
+                                    height: '100%',
+                                }}
+                            >
+                                <Loader overlay />
+                            </div>
+                        }
+                        <div style={{height: '100%'}}>
+                            <Panel background='light' open={this.state.previewItem != null} side='right' size="medium">
```
MarkLark86, line 282: Does specifying `background='light'` work with a dark theme?

client/extensions/relatedArticles/src/EventsRelatedArticles/PreviewArticle.tsx
```diff
@@ -0,0 +1,131 @@
+import React from 'react';
```
MarkLark86, line 1: We already have an [ArchivePreviewComponent](https://github.com/superdesk/superdesk-planning/blob/develop/client/components/Archive/ArchivePreview/index.tsx), which is used in the Assignment side panel. Could this be used here? They both provide the same functionality, so it's best to stick to DRY principle of re-using an existing component

client/extensions/relatedArticles/src/EventsRelatedArticles/RelatedArticlesListComponent.tsx
```diff
@@ -0,0 +1,188 @@
+import React from 'react';
```
MarkLark86, line 1: Same thing here, stick with the DRY principle and use `ArticleItemConcise` component from the SuperdeskAPI.
This way this one component is used in multiple places to render related articles, rather than duplication of the same functionality.

If we need changes done, such as to show language, then this should be done on the `ArticleItemConcise` component

client/extensions/relatedArticles/src/EventsRelatedArticles/utils.ts
```diff
@@ -0,0 +1,196 @@
+import {IArticle} from 'superdesk-api';
+
+export const fakeEditor: any = {
```
MarkLark86, line 3: I'm assuming this is not needed for production, only in testing/development. If so, please remove this

thecalcc: We do need it, because the items coming from Belga don't have an actual content profile, but the `search_provider_proxy` from our backend works with projections, and those are gotten through a content profile. In the future I guess we can rework this so we don't need to fake content profiles.

thecalcc:
Moved to superdesk-planning (https://github.com/superdesk/superdesk-planning/pull/1931)

PR #444, opened 2024-04-08
Add new belga specific assignment notifications templates [SDBELGA-781]

devketanpro:


server/templates/assignment_details_email.html
```diff
@@ -0,0 +1,61 @@
+git<h2>Planning Item: {{ slugline }}</h2>
```
petrjasek: `git`

server/templates/assignment_details_email.html
```diff
@@ -0,0 +1,61 @@
...
+    {% if event.get('slugline') %}
+        <h3>Event: {{ event['slugline'] }}</h3>
+    {% endif %}
+
+    <p><strong>Event Url:</strong> <a href="{{ client_url }}/#/planning?preview={'id': '{{ event['id'] }}','type':'event' }}">View Event Details</a></p>
+
+    {% if event.get('calendars') %}
+        <p><strong>Calendars:</strong> 
+        {% for calendar in event.get('calendars', []) %}
+            {{ calendar.get('name', '') }},
```
petrjasek: you can do this via:
```
{{ event.calendars|map(attribute="name")|join(", ") }}
```
that will avoid that trailing `,`

server/templates/assignment_details_email.html
```diff
@@ -0,0 +1,61 @@
...
+
+    {% if event.get('name') %}
+        <p><strong>Event Name:</strong> {{ event['name'] }}</p>
+    {% endif %}
+
+    {% if event.get('definition_short') %}
+        <p><strong>Description:</strong> {{ event['definition_short'] }}</p>
+    {% endif %}
+
+    {% if contacts | length is gt 0 %}
```
petrjasek: `if contacts` would be enough, empty list is like `False`

petrjasek (approved):

PR #447, opened 2024-04-19
Introduce Planning export templates in Dutch and French [SDBELGA-784]

devketanpro:


server/tests/planning_export/planning_export.py
```diff
@@ -0,0 +1,309 @@
...
+                        "scheme": "belga-keywords",
+                        "translations": {
+                            "name": {"nl": "MUSIC FESTIVALS", "fr": "MUSIC FESTIVALS"}
+                        },
+                    }
+                ],
+            },
+        ]
+
+        data = format_event_french(events)
```
petrjasek: would be good to test it using `render_template` and not just the helper

server/belga/planning_exports/france_news_event_list.py
```diff
@@ -0,0 +1,49 @@
...
+    end_date = format_date(
+        sorted_events[-1].get("dates").get("start"), "EEEE d", locale="fr"
+    ).capitalize()
+    month = format_date(
+        sorted_events[0].get("dates").get("start"), "MMMM", locale="fr"
+    ).capitalize()
+
+    intro_text = {
+        "title": (
+            f"Calendrier sportif international du "
```
petrjasek: this text handling would be better in the template so it's easier to change it

server/templates/dutch_news_events_list_export.html
```diff
@@ -0,0 +1,23 @@
+{% set formatted_data = format_event_dutch(items) %}
+<div>
+    <h2><b>{{ formatted_data.intro.title }}</b></h2>
```
petrjasek: that `b` inside `h2` seems not very useful

server/templates/dutch_news_events_list_export.html
```diff
@@ -0,0 +1,23 @@
...
+<div>
+    {% for event in day_events.events %}
+    <div>
+        <p>{{ event.subject }}<br></p>
+        <p>{{ event.address.name }}, {{event.address.country}} <br></p>
+        <p>{{ event.local_time }}<br></p>
+        <p>{{ event.title }} <br></p>
+        <p>{{ event.description }}<br></p>
+        {% for link in event.links %}
+        <div><a href="{{ link }}">Link</a><br></div>
```
petrjasek: imo might be better to show the actual link instead of `Link` text

server/templates/dutch_news_events_list_export.html
```diff
@@ -0,0 +1,23 @@
...
+<div>
+    <h2>Internationale sportkalender van {{formatted_data.start_date}} tot {{formatted_data.end_date}} {{formatted_data.month}}</h2>
+    <p>De belangrijkste sportevenementen op de Belgische en internationale sportkalender van {{formatted_data.start_date}} tot {{formatted_data.end_date}} {{formatted_data.month}}:</p>
+</div>
+{% for day_events in formatted_data.events_list %}
+    <h3>{{ day_events.date }}</h3>
+<div>
+    {% for event in day_events.events %}
+    <div>
+        <p>{{ event.subject }}<br></p>
```
petrjasek: I wonder if all those `<br>`s are needed if each line is it's own `p` tag

server/templates/dutch_news_events_list_export.html
```diff
@@ -0,0 +1,23 @@
...
+        <p>{{ event.address.name }}, {{event.address.country}} <br></p>
+        <p>{{ event.local_time }}<br></p>
+        <p>{{ event.title }} <br></p>
+        <p>{{ event.description }}<br></p>
+        {% for link in event.links %}
+        <div><a href="{{ link }}">{{ link }}</a><br></div>
```
petrjasek: might be better to use `p` here instead of `div` to have consistent lines

petrjasek (approved):

PR #452, opened 2024-05-06
implement belga specific coverage due date callback

petrjasek:
SDBELGA-782

uses https://github.com/superdesk/superdesk-planning/pull/1981

### DIFF
--- a/client/index.tsx
+++ b/client/index.tsx
@@ -10,7 +10,7 @@ import belga360Archive from './belga/360archive';
 import belgaPress from './belga/belgapress';
 import moment from 'moment';
 import {IEventItem, IPlanningItem} from 'superdesk-planning/client/interfaces';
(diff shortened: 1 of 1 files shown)


MarkLark86 (approved):

client/index.tsx
```diff
@@ -8,6 +8,10 @@ import {AvatarContentText} from 'superdesk-ui-framework';
 import belgaImage from './belga/image';
 import belga360Archive from './belga/360archive';
 import belgaPress from './belga/belgapress';
+import moment from 'moment';
+import {IEventItem, IPlanningItem} from 'superdesk-planning/client/interfaces';
```
MarkLark86, line 12: For now this should be fine, but ideally we should have a public interface for the Planning module that would provide these utils and configuration abilities.

petrjasek: yep didn't want to complicate it now but some public interface would be the way to go

PR #499, opened 2024-06-18
Fix : Sports Advisory custom layout issues [SDBELGA-822]

devketanpro:


server/belga/command/planning_export_templates_update.py
petrjasek, line 1: what's the usecase for this file?

petrjasek:
why not just update `data/planning_export_templates.json` ? it might require some core changes to allow `init_version` but would be easier long term instead of writing migrations for each change

petrjasek (approved):

PR #503, opened 2024-06-20
FIX : Daily advisory custom layout issues [SDBELGA-829]

devketanpro:
Issues fix:-

1. Locations - show the full address not only Country and City.
2. Show only those text Coverages which match the language of the custom layout
3. In the French custom layout, the items show up in Dutch if they are available in two languages (make sure only the French version is shown)

### DIFF
--- a/server/belga/planning_exports/common.py
+++ b/server/belga/planning_exports/common.py
@@ -53,7 +53,7 @@ def set_item_dates(item: Dict[str, Any], event: Dict[str, Any]):
     item["local_date_time"] = start_local.strftime("%Y%m%d")
 
 
-def set_item_location(event: Dict[str, Any]):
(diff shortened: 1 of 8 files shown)


server/belga/planning_exports/common.py
```diff
@@ -53,21 +53,29 @@ def set_item_dates(item: Dict[str, Any], event: Dict[str, Any]):
     item["local_date_time"] = start_local.strftime("%Y%m%d")
 
 
-def set_item_location(item: Dict[str, Any], event: Dict[str, Any]):
+def set_item_location(event: Dict[str, Any]):
```
petrjasek: this should be probably called `get_item_location` and you can add return type

petrjasek (approved):

PR #515, opened 2024-07-04
fix: check why many DPA texts are not getting ingested [SDBELGA-844]

devketanpro:


server/belga/io/feed_parsers/belga_dpa_newsml_2_0.py
```diff
@@ -205,7 +205,8 @@ def parse_content_subject(self, tree, item):
                     item.setdefault('extra', {})['country'] = name_elt.text
                 # looking for ISO 3166-1 alpha-3 code
                 for i in subject_elt.findall(self.qname('sameAs')):
-                    code = i.find(self.qname('name')).text
+                    name_as_elt = i.find(self.qname('name'))
+                    code = name_as_elt.text if name_as_elt is not None else i.get("qcode").split(':')[1]
```
petrjasek: imo this should be rather:
```py
code = i.get("qcode").split(":")[1] if i.get("qcode").startswith("iso3166-1a3") else ""

Comment (petrjasek) [reply server/belga/io/feed_parsers/belga_dpa_newsml_2_0.py]: so it won't be using any random 3 letter code they might introduce later

Comment (devketanpro) [reply server/belga/io/feed_parsers/belga_dpa_newsml_2_0.py]: got it

File: server/belga/io/feed_parsers/belga_dpa_newsml_2_0.py [hunk 2]
```diff
@@ -226,9 +226,9 @@ def parse_content_subject(self, tree, item):
                 for i in subject_elt.findall(self.qname("sameAs")):
                     name_as_elt = i.find(self.qname("name"))
                     code = (
-                        name_as_elt.text
-                        if name_as_elt is not None
-                        else i.get("qcode").split(":")[1]
+                        i.get("qcode").split(":")[1]
+                        if i.get("qcode").startswith("iso3166-1a3")
+                        else name_as_elt.text
```
petrjasek: this could be still None, not sure why do we need it?

devketanpro: yo, my bad

petrjasek (approved):

PR #518, opened 2024-07-08
add new prefrences for assignment and mark_for_user notications [SDBELGA-818]

devketanpro:


petrjasek:
why is this only for belga?

devketanpro:
> why is this only for belga?

Since this ticket is related to Belga, I assume it is specific to Belga.

devketanpro:
Its a core feature, not for Belga specific, so closing the PR

PR #592, opened 2024-09-12
Fix comma between postalcode and city in the daily advisories [SDBELGA-870]

devketanpro:


server/belga/planning_exports/common.py
```diff
@@ -197,9 +196,11 @@ def get_coverages(event: Dict[str, Any], locale: str):
                 coverage.get("news_coverage_status", {}).get("label", "").upper()
             )
 
-            formatted_coverages.append(
-                f"{cov_type} ({cov_status})"
-            ) if cov_planning.get("language", locale) == locale else []
+            (
+                formatted_coverages.append(f"{cov_type} ({cov_status})")
+                if cov_planning.get("language", locale) == locale
+                else []
```
petrjasek: this would be better readable like:
```py
if cov_planning.get("language", locale) == locale:
    formatted_coverages.append(f"{cov_type} ({cov_status})")
```
or even on top you could do:
```py
if cov_planning.get("language", locale) != locale:
    continue
```

petrjasek (approved):

PR #618, opened 2024-10-08
update analytics

devketanpro:


petrjasek:
I think we need it in superdesk-tga repo rather than here

PR #622, opened 2024-10-16
update core to release/2.8

devketanpro:


devketanpro:
Hey @petrjasek, I tried compiling the requirements file with both Python 3.8 and 3.10, but I still encountered dependency conflicts. Can you please take a look at it when you get a chance?

Thanks!

petrjasek:
pushed an update that seems working, seems like you have to do it on 3.10 but still required some tweaking

PR #673, opened 2025-02-05
Fix Update Event Group dates instead of headlines [SDBELGA-934]

devketanpro:


server/tests/planning_export/planning_export_tests.py
```diff
@@ -526,7 +526,7 @@ def test_export_week(self):
                 ),
                 dutch_template_data,
             )
-            self.assertIn("<h3>zondag 21 april</h3>", dutch_template_data)
+            self.assertIn("<h3>Zondag 21 april</h3>", dutch_template_data)
```
petrjasek, line 529: what about `April` that shouldn't be capitalized?

devketanpro: not sure, as they only want the first letter of the heading to be capitalized.

petrjasek (approved):

PR #698, opened 2025-05-01
Translations of published agency ECO-texts should keep the similar package [EXT/ECO (fr) = BTL/ECO (nl) and vice versa] [SDBELGA-720]

devketanpro:


server/belga/macros/set_default_metadata_with_translate.py
```diff
@@ -18,6 +18,41 @@
...
+
+    # filter out any existing services-products entries
+    filtered = [
+        s for s in new_item.get("subject", []) if s.get("scheme") != "services-products"
+    ]
+
+    # build the new subject entry
+    filtered.append(
+        {
+            "name": mapped,
```
petrjasek: would rather populated it based on the values from CV and not hardcode it

petrjasek (approved):

PR #759, opened 2025-08-26
As a Belga editor I would like to see and handle the videos provided by picture pack in Superdesk [SDBELGA-901]

devketanpro:
image


server/settings.py
```diff
@@ -357,6 +357,8 @@
 
 BELGA_AI_URL = env("BELGA_AI_URL")
 
+BELGA_VIDEO_ENABLED = True
```
petrjasek: we would need env variable to control it without releasing

server/belga/search_providers.py
```diff
@@ -240,6 +240,78 @@ def url(self, resource):
...
+        data = self.api_get(self.search_endpoint, api_params)
+        docs = [self.format_list_item(item) for item in data[self.items_field]]
+        return BelgaListCursor(docs, data[self.count_field])
+
+    def format_list_item(self, data):
+        """Format list items. Only override for videos, delegate pictures to parent."""
+
+        item_id = str(data.get("ID") or data.get("imageId") or "")
+        is_video = (
+            data.get("ObjectType") == 1
```
petrjasek: isn't the type check enough?

server/belga/search_providers.py
```diff
@@ -240,6 +240,78 @@ def url(self, resource):
...
+            .startswith("video")
+        )
+
+        if not is_video:
+            return super().format_list_item(data)
+
+        guid = "urn:belga.be:picturepackmedia:%s" % item_id
+        created = get_datetime(data.get("EntryDate", data.get("createDate")))
+
+        return {
```
petrjasek: seems like there is lots of code duplicated, would be good to reuse it

server/belga/search_providers.py
```diff
@@ -240,6 +240,78 @@ def url(self, resource):
         base_url = config.get("url") or self.base_url
         return urljoin(base_url, resource.lstrip("/"))
 
+    def find(self, query, params=None):
+        api_params = {
+            "s": query.get("from", 0),
+            "l": query.get("size", 25),
+        }
+
+        if params and app.config.get("BELGA_VIDEO_ENABLED", False):
```
petrjasek: I guess when video is not enabled we should add filter param to only get images

server/belga/search_providers.py
```diff
@@ -240,6 +240,57 @@ def url(self, resource):
...
+
+        item.update(
+            {
+                "_id": f"urn:belga.be:picturepackmedia:{data.get('ID') or data.get('imageId')}",
+                "guid": f"urn:belga.be:picturepackmedia:{data.get('ID') or data.get('imageId')}",
+                "type": "video",
+                "mimetype": "video/mp4",
+                "renditions": {
+                    "thumbnail": {
+                        "href": data.get("ThumbnailURL", data.get("thumbnailUrl", "")),
```
petrjasek, line 276: pls add mimetype here

petrjasek (approved):

PR #771, opened 2025-09-02
Introduce advisory layouts in Dutch and French for Planning items filtered by Coverage type [SDBELGA-973]

devketanpro:


server/belga/planning_exports/format_planning_for_tomorrow.py
```diff
@@ -0,0 +1,92 @@
...
+            "location": get_item_location(event, locale),
+            "links": event.get("links", []),
+        }
+        set_metadata(formatted_event, event, locale)
+
+        dates = formatted_event["dates"]
+        start_local = utc_to_local(dates["tz"], dates["start"])
+        end_local = utc_to_local(dates["tz"], dates["end"])
+        formatted_event["time"] = (
+            f"{start_local.strftime('%H:%M')} - {end_local.strftime('%H:%M')}"
```
petrjasek, line 79: do they use any all day events?

server/templates/dutch_planning_advisory_tomorrow.html
```diff
@@ -0,0 +1,31 @@
...
+<div>
+    <h2>Dit is de Belga-agenda van de Belgische en internationale gebeurtenissen, met de vermelding of wij dit in TEXT, PICTURE, VIDEO, AUDIO, INFOGRAPHICS, LIVE VIDEO en LIVE BLOG coveren. De vermelding ON MERIT betekent dat Belga dit onderwerp opvolgt, maar dat voorlopig niet gegarandeerd kan worden dat er ook een specifieke coverage zal volgen. De redactie van Belga wenst u een prettige werkdag.</h2>
+</div>
+
+{% for data in formatted_data %}
+    <h3>{{ data.calendar }}</h3>
+    <div>
+        {% for event in data.events %}
+            <div>
+                <p>{{ event.time }}<br></p>
```
petrjasek: not sure why are there those line breaks everywhere? wouldn't `p` handle it?

server/belga/planning_exports/format_planning_for_tomorrow.py
devketanpro: we don’t currently use all-day events in our implementation, including the existing format_event_for_tommorow. The formatting has always been based on start and end times (HH:MM - HH:MM). If required, we can extend it to detect and label all-day events, but this was never part of the original functionality.

petrjasek (approved):

PR #797, opened 2025-09-24
Fix the issues with the Picture Pack search provider [SDBELGA-992]

devketanpro:


server/belga/search_providers.py
```diff
@@ -246,11 +246,39 @@ def find(self, query, params=None):
...
+                )
+            if dates.get("end"):
+                api_params["e"] = int(
+                    arrow.get(dates["end"], "DD/MM/YYYY").timestamp() * 1000
+                )
+
+            if params.get("period"):
+                api_params["p"] = params["period"].upper()
+
+            if app.config.get("BELGA_VIDEO_ENABLED", False) and params.get(
```
petrjasek: can you also apply `strtobool` in the settings and make it off by default

petrjasek (commented):
there is lot of code duplicated now from the original fetch, would be good to move that param parsing into a method which could be extended in the picture pack instead of doing the whole fetch there

server/settings.py
```diff
@@ -357,8 +358,7 @@
 
 BELGA_AI_URL = env("BELGA_AI_URL")
 
-BELGA_VIDEO_ENABLED = env("BELGA_VIDEO_ENABLED", True)
-
+BELGA_VIDEO_ENABLED = bool(strtobool(env("BELGA_VIDEO_ENABLED", "True")))
```
petrjasek, line 361: pls set it to `False` by default

server/settings.py
```diff
@@ -357,8 +358,7 @@
 
 BELGA_AI_URL = env("BELGA_AI_URL")
 
-BELGA_VIDEO_ENABLED = env("BELGA_VIDEO_ENABLED", True)
-
+BELGA_VIDEO_ENABLED = bool(strtobool(env("BELGA_VIDEO_ENABLED", "False")))
```
petrjasek: I think there is no need for that `bool`

petrjasek (approved):

petrjasek (approved):

PR #798, opened 2025-09-25
Daily program in two languages for the Event and Planning advisories [SDBELGA-979]

devketanpro:
Fix Planning Item Export for Bilingual Advisory

### DIFF
--- a/server/tests/planning_export/planning_export_tests.py
+++ b/server/tests/planning_export/planning_export_tests.py
@@ -869,7 +869,7 @@ class PlanningExportTests(TestCase):
             )
 
     def test_bilingual_planning_advisory_export_tomorrow(self):
-        """Test the bilingual planning advisory template with various scenarios"""
+        """Test the bilingual planning advisory template using only planning data"""
         with self.app.app_context():
             test_contact_id = ObjectId()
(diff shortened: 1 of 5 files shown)


server/tests/planning_export/planning_export_tests.py
```diff
@@ -896,15 +896,18 @@ def test_bilingual_planning_advisory_export_tomorrow(self):
...
                 {
-                    "_id": event_1_id,
+                    "_id": planning_1_id,
+                    "type": "planning",
+                    "slugline": "planning-dutch-text",
                     "name": "Test Event NL",
-                    "slugline": "test-event-nl",
-                    "definition_long": "Dutch description of the event",
+                    "description_text": "Dutch description of the event",
                     "dates": {
```
petrjasek: I suppose it's not used, but it wouldn't be in planning items so better remove it so it's not confusing

devketanpro: Thanks Petr, actually description_text is used in planning items

petrjasek: I mean `dates`

petrjasek (approved):

PR #807, opened 2025-10-03
Daily program in two languages for the Event and Planning advisories - For internal use [SDBELGA-990]

devketanpro:


petrjasek (commented):
pls add test to make sure those templates render

petrjasek (approved):

PR #850, opened 2025-12-09
Daily advisories - Bugs [SDBELGA-1007]

devketanpro:


server/belga/planning_exports/internal_bilingual_events_advisory_tomorrow.py
```diff
@@ -30,11 +30,24 @@ def format_event_for_tommorow_bilingual_internal(
 
     # Determine weekday and date from the event
     if event_data:
-        first_event_start = utc_to_local(
-            event_data[0]["dates"].get("tz", "Europe/Brussels"),
-            event_data[0]["dates"]["start"],
-        )
-        weekday_date = first_event_start.strftime("%A %d %B %Y").upper()
+        local_dates = []
```
petrjasek: this should go to some utils and be reused in both files

server/belga/planning_exports/format_planning_for_tomorrow_bilingual.py
```diff
@@ -219,6 +219,6 @@ def get_advisory_weekday_date(planning_item, planning_service=None, event_servic
 
     if scheduled:
         local_dt = utc_to_local("Europe/Brussels", scheduled)
-        return local_dt.strftime("%A %d %B %Y").upper()
+        return local_dt.strftime("%A %-d %B %Y").upper()
```
petrjasek: would be good to have a constant for it and not fixing it in multiple places

server/belga/planning_exports/internal_bilingual_events_advisory_tomorrow.py
petrjasek: this section still looks duplicated

petrjasek (approved):

PR #985, opened 2026-06-12
As a Belga editor I want to get some French and Dutch news automatically translated into English and pushed to a specific Desk [SDBELGA-1100]

devketanpro:


server/belga/macros/auto_translate_to_en.py
```diff
@@ -159,11 +159,10 @@ def translate_item(item):
 def auto_translate_to_en(item, **kwargs):
     try:
         translate_item(item)
-    except Exception as exc:
-        logger.error(
-            "Auto translate to en failed for item %s: %s",
+    except Exception:
+        logger.exception(
+            "Auto translate to en failed for item %s",
```
petrjasek: would avoid item specific exception, make the item an extra param

petrjasek (approved):

PR #995, opened 2026-07-02
[SDBELGA-1109] fix: Adjust coverage due dates strategy to be timezone aware

BrianMwangi21:
### Purpose
This updates the Belga coverage due-date strategy to evaluate all-day and TBC events in the event's own timezone instead of the browser timezone.

### DIFF
--- a/client/index.tsx
+++ b/client/index.tsx
@@ -8,10 +8,8 @@ import {AvatarContentText} from 'superdesk-ui-framework';
 import belgaImage from './belga/image';
 import belga360Archive from './belga/360archive';
 import belgaPress from './belga/belgapress';
-import moment from 'moment';
-import {IEventItem, IPlanningItem} from 'superdesk-planning/client/interfaces';
 import {setCoverageDueDateStrategy} from 'superdesk-planning/client/configure';
-import {eventUtils, timeUtils} from 'superdesk-planning/client/utils';
+import {getCoverageDueDate} from './utils/coverageDueDate';
 
 class UserAvatar extends React.PureComponent<{user: Partial<IUser>}> {
     render() {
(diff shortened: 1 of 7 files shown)


client/index.tsx
```diff
@@ -38,23 +38,35 @@ function getCoverageDueDate(
...
+        timeUtils.getDateInRemoteTimeZone(eventItem.dates.start, eventTimezone) :
+        moment(planningItem?.planning_date);
+    const endInEventTimezone = eventItem?.dates?.end && eventTimezone ?
+        timeUtils.getDateInRemoteTimeZone(eventItem.dates.end, eventTimezone) :
+        moment(planningItem?.planning_date);
+    const isAllDayEvent = eventItem?.dates?.all_day || (
+        !!eventItem?.dates?.start &&
+        !!eventItem?.dates?.end &&
+        eventUtils.isEventAllDay(startInEventTimezone, endInEventTimezone, true)
+    );
```
eos87: This is kinda hard to read and also there is some duplication we could get rid by extracting it into a tiny helper like 

```typescript
const getEventDateContext = (eventItem?: IEventItem, planningItem?: IPlanningItem) => {
    const tz = eventItem?.dates?.tz;
    const fallback = moment(planningItem?.planning_date);

    const toEventTz = (value?: string) =>
        value && tz ? timeUtils.getDateInRemoteTimeZone(value, tz) : fallback.clone();

    const start = toEventTz(eventItem?.dates?.start);
    const end = toEventTz(eventItem?.dates?.end);

    const isAllDay =
        !!eventItem?.dates?.all_day ||
        (!!eventItem?.dates?.start &&
            !!eventItem?.dates?.end &&
            eventUtils.isEventAllDay(start, end, true));

    return {start, end, isAllDay};
};
```

client/index.tsx
```diff
@@ -38,23 +38,35 @@ function getCoverageDueDate(
...
     } else if (eventItem) {
-        coverageTime = moment(eventItem.dates?.end);
+        coverageTime = endInEventTimezone.clone();
         coverageTime.add(1, 'hour');
-        if (eventItem.dates?.end && !coverageTime.isSame(eventItem.dates?.end, 'day')) {
+        if (eventItem.dates?.end && !coverageTime.isSame(endInEventTimezone, 'day')) {
             // make sure we're not going into the next day
-            coverageTime = moment(eventItem.dates?.end);
+            coverageTime = endInEventTimezone.clone();
         }
```
eos87: it would be good to extract also this and cover it with proper unit tests to "future-proof" it

eos87 (approved):
Nicely done! Thanks.

